PR #9, opened 2023-08-16
code and data for simulation of brain phantom

oscarjalnefjord:
A simple simulation based on a digital brain atlas. To be used until IVIM phantoms can be made with the general DRO that TF 3.1 is working on.


phantoms/brain/sim_brain_phantom.py
```diff
@@ -0,0 +1,53 @@
+import os
+import shutil
+import json
+import numpy as np
+import nibabel as nib
+from scipy.ndimage import zoom
+
+regime = 'diffusive'#'ballistic' #
```
etpeterson, line 8: Maybe check if the `regime` is strictly either `diffusive` or `ballistic`. Right now a simple misspelling like `difusive` would actually generate the ballistic phantom.

oscarjalnefjord: Good point. I use predefined string variables now instead. Robust enough?

PR #12, opened 2023-09-18
Testing/wrapper

IvanARashid:
Branch containing wrapper and basic method for mass-testing algorithms implemented according to OsipiBase.

The "standardized" algorithms are contained in the src/standardized folder. Apart from identification attributes and requirements, they consist of a single ivim_fit method which performs the fit according to the algorithm. These inherit from OsipiBase which is our default object that will be used for interaction. To run a fit, the OsipiBase.osipi_fit method should be called. This allows for mass-testing later.

The testing is performed in tests/unit_tests/test_ivim_fit.py


DKuppens:
Can we make the tests in tests/unit_tests/test_ivim_fit.py pass? Currently 2 fail.
Also, how is tests/unit_tests/simple_test_run_of_algorithm.py supposed to be used? Currently I get a fixture error if I run pytest in that file.

DKuppens:
Is it intentional to add the bytecode to the repo?

src/wrappers/OsipiBase.py
```diff
@@ -0,0 +1,210 @@
...
+        This method can be used instead of specifically importing that class and initiating it.
+        
+        WIP: Add args and kwargs!
+
+        Args:
+            algorithm (string): The name of the algorithm, should be the same as the file in the src/standardized folder without the .py extension.
+        """
+        
+        # Import the algorithm
+        import_base_path = "src.standardized"
```
DKuppens, line 32: Does this always work or only when the current working directory is equal to the root dir of this repo? Maybe we should check if the root dir is added to pythonpath then?

DKuppens:
What is the difference between the ivim_fit and osipi_fit methods?

src/wrappers/OsipiBase.py
IvanARashid: Not sure, haven't tested that. I'm always in the root directory of the folder. I could see if I can add a check for that. Do you have an idea of what we could do if it is not in pythonpath? Or do we just raise an error, asking people to be in the root dir?

IvanARashid:
> Can we make the tests in tests/unit_tests/test_ivim_fit.py pass? Currently 2 fail. Also, how is tests/unit_tests/simple_test_run_of_algorithm.py supposed to be used? Currently I get a fixture error if I run pytest in that file.

I believe the two that fail is due to flaws in the algorithm. For example, if you uncomment the lines that have to do with negative values, we only get one fail. In the results, we can see that it is the ETP algorithm that fails, while the IAR alorithm doesn't. The difference between the two is that ETP does a linear fit, while IAR is a simple bi-exponential NLLS fit. So if IAR passes, then it's probably something with the linear fit that can't handle those specific testing parameters. You can try to play around with the tolerances and see if you can get ETP to pass. In the end, I think Eric knows more about the issue than I do.

Regarding simple_test_run_of_algorithms, this is a simple "run" script, which initiates and performs a single fit for the algorithms implemented in OsipiBase. It was used when I wrote the standardized versions, just to see if they worked. I added it to the repo as it could be useful for doing similar test runs when we implement the rest of the submissions according to OsipiBase. So it is not supposed to be ran with pytest.

IvanARashid:
> Is it intentional to add the bytecode to the repo?

Sorry about that, though I had them in git ignore. I think I've removed all of them now

IvanARashid:
> What is the difference between the ivim_fit and osipi_fit methods?

My thinking is that osipi_fit should include checks, where inputs are compared to algorithm requirements defined as attributes in the algorithm subclass. So instead of calling the checks in all ivim_fit methods we write, they are instead called from the parent class. So it's more for future functionality.

Also, we may develop the testing further. Test for different fit settings, different dimensions, etc. By having a fit method that performs checks, we can quite practically just loop over them in testing, and run tests on methods that can take the inputs we provide.

src/wrappers/OsipiBase.py
etpeterson: What is the advantage of this dynamic loading vs hard coding all the available algorithms into a file somewhere (or even this file) and referencing it that way?

etpeterson:
> > Can we make the tests in tests/unit_tests/test_ivim_fit.py pass? Currently 2 fail. Also, how is tests/unit_tests/simple_test_run_of_algorithm.py supposed to be used? Currently I get a fixture error if I run pytest in that file.
> 
> I believe the two that fail is due to flaws in the algorithm. For example, if you uncomment the lines that have to do with negative values, we only get one fail. In the results, we can see that it is the ETP algorithm that fails, while the IAR alorithm doesn't. The difference between the two is that ETP does a linear fit, while IAR is a simple bi-exponential NLLS fit. So if IAR passes, then it's probably something with the linear fit that can't handle those specific testing parameters. You can try to play around with the tolerances and see if you can get ETP to pass. In the end, I think Eric knows more about the issue than I do.
> 
> Regarding simple_test_run_of_algorithms, this is a simple "run" script, which initiates and performs a single fit for the algorithms implemented in OsipiBase. It was used when I wrote the standardized versions, just to see if they worked. I added it to the repo as it could be useful for doing similar test runs when we implement the rest of the submissions according to OsipiBase. So it is not supposed to be ran with pytest.

Regarding the failing tests, I haven't looked closely at the code yet, but I think there are a few things we need to figure out.
First, I'm a little suspicious of the failures because isn't the same test data passing in `main`?
Regardless of the details though, I think this is showing that not all algorithms can handle all data to the same degree and we need to be flexible about that. I think the best approach would be to have some default test values for each algorithm that we can load and use for testing. That would preserve the unit testing we want to do and also allow customized bounds (or other customizations) for each algorithm. Failing that I think we'd just need to be too permissive on the testing side to generate data and the value of the testing becomes basically nothing.
I'm also OK just removing the linear fitting from the list to be tested and replacing it with another, more capable algorithm. IMO it would be interesting to include as a baseline reference algorithm, but if it's causing too much trouble then let's take the easy path for now.

src/wrappers/OsipiBase.py
```diff
@@ -0,0 +1,210 @@
...
+        self.__init__(**kwargs)
+
+    def osipi_fit(self, data=None, bvalues=None, thresholds=None, bounds=None, initial_guess=None, **kwargs):
+        """Fits the data with the bvalues
+        Returns [S0, f, Dstar, D]
+        """
+        
+        # We should first check whether the attributes in the __init__ are not None
+        # Then check if they are input here, if they are, these should overwrite the attributes
+        use_bvalues = bvalues if bvalues is not None else self.bvalues
```
etpeterson, line 46: `np.asarry` for these?

src/wrappers/OsipiBase.py
```diff
@@ -0,0 +1,210 @@
...
+        # Then we pass everything into the ivim_fit method which performs the fit according to the algorithm
+        
+        #args = [data, use_bvalues, use_initial_guess, use_bounds, use_thresholds]
+        args = [arg for arg in args if arg is not None]
+        results = self.ivim_fit(*args, **kwargs)
+        
+        #self.parameter_estimates = self.ivim_fit(data, bvalues)
+        return results
+    
+    def osipi_print_requirements(self):
```
etpeterson, line 73: Take it or leave it, but this looks a lot like a `__str__` function. Maybe having that would be useful too?

src/wrappers/OsipiBase.py
```diff
@@ -0,0 +1,210 @@
...
+        (True, True, False, False, False, False)
+        """
+        
+        return (False,) * 6
+
+    def osipi_accepts_dimension(self, dim):
+        """Query if the selection dimension is fittable"""
+        
+        accepted = self.accepted_dimensions()
+        if dim < 0 or dim > len(accepted):
```
etpeterson, line 132: I think there's an array offset here in the two getters. The general getter, `osipi_accepted_dimensions` seems to indicate that the array starts at 1D and goes to 6D. But the specific getter, `osipi_accepts_dimension` doesn't have the offset, so `accepted[0]` looks up the first value, which should reference the 1D case.
Perhaps the accepted dimensions should be stored as a dictionary with the dimension as the key and `True` or `False` as the payload?

src/wrappers/OsipiBase.py
IvanARashid: Manual work :) Just the possibility of forgetting to add one really. We already need to implement them according to our standard and in addition to that, add them to a registry of submissions. A third place/thing to do seems excessive. However, depending on how we do the registry, we could also just get the names from there. Of course, this is me assuming that we will drown in submissions in the future. And the chance of that happening seems quite slim at the moment.

src/wrappers/OsipiBase.py
```diff
@@ -0,0 +1,210 @@
...
+            return False
+        return accepted[dim]
+    
+    def osipi_check_required_bvalues(self):
+        """Checks if the input bvalues fulfil the algorithm requirements"""
+        
+        if self.bvalues.size < self.required_bvalues:
+            print("Conformance error: Number of b-values.")
```
etpeterson, line 140: How do you see this working? Right now it's a print and a return value. But I also could see it asserting that the sizes are accurate. I suppose it depends how strict we want to be.

src/wrappers/OsipiBase.py
```diff
@@ -0,0 +1,210 @@
...
+        if self.required_initial_guess is True and self.initial_guess is None:
+            print("Conformance error: Initial guess")
+            return False
+        else:
+            return True
+
+    
+    def osipi_check_required_bvalues():
+        """Minimum number of b-values required"""
+        pass
```
etpeterson, line 174: The other "checks" return `True` or `False`. Perhaps this template should just return `True`?

src/wrappers/OsipiBase.py
```diff
@@ -0,0 +1,210 @@
...
+
+    def osipi_accepts_dimension(self, dim):
+        """Query if the selection dimension is fittable"""
+        
+        accepted = self.accepted_dimensions()
+        if dim < 0 or dim > len(accepted):
+            return False
+        return accepted[dim]
+    
+    def osipi_check_required_bvalues(self):
```
etpeterson, line 136: Unless GitHub is missing it, these checks aren't being run. Should they be?

IvanARashid:
> > > Can we make the tests in tests/unit_tests/test_ivim_fit.py pass? Currently 2 fail. Also, how is tests/unit_tests/simple_test_run_of_algorithm.py supposed to be used? Currently I get a fixture error if I run pytest in that file.
> > 
> > 
> > I believe the two that fail is due to flaws in the algorithm. For example, if you uncomment the lines that have to do with negative values, we only get one fail. In the results, we can see that it is the ETP algorithm that fails, while the IAR alorithm doesn't. The difference between the two is that ETP does a linear fit, while IAR is a simple bi-exponential NLLS fit. So if IAR passes, then it's probably something with the linear fit that can't handle those specific testing parameters. You can try to play around with the tolerances and see if you can get ETP to pass. In the end, I think Eric knows more about the issue than I do.
> > Regarding simple_test_run_of_algorithms, this is a simple "run" script, which initiates and performs a single fit for the algorithms implemented in OsipiBase. It was used when I wrote the standardized versions, just to see if they worked. I added it to the repo as it could be useful for doing similar test runs when we implement the rest of the submissions according to OsipiBase. So it is not supposed to be ran with pytest.
> 
> Regarding the failing tests, I haven't looked closely at the code yet, but I think there are a few things we need to figure out. First, I'm a little suspicious of the failures because isn't the same test data passing in `main`? Regardless of the details though, I think this is showing that not all algorithms can handle all data to the same degree and we need to be flexible about that. I think the best approach would be to have some default test values for each algorithm that we can load and use for testing. That would preserve the unit testing we want to do and also allow customized bounds (or other customizations) for each algorithm. Failing that I think we'd just need to be too permissive on the testing side to generate data and the value of the testing becomes basically nothing. I'm also OK just removing the linear fitting from the list to be tested and replacing it with another, more capable algorithm. IMO it would be interesting to include as a baseline reference algorithm, but if it's causing too much trouble then let's take the easy path for now.

I think it's good to see some fails. It means that our tolerances are working. I don't know how it affects the results though. Can we still log the estimates, etc?

tests/IVIMmodels/unit_tests/simple_test_run_of_algorithm.py
```diff
@@ -0,0 +1,29 @@
+import numpy as np
+import os
+from pathlib import Path
+from src.standardized.ETP_SRI_LinearFitting import ETP_SRI_LinearFitting
+from src.standardized.IAR_LU_biexp import IAR_LU_biexp
+
+## Simple test code... 
+# Used to just do a test run of an algorithm during development
+def test_run(model, **kwargs):
```
etpeterson, line 9: I think pytest is confused by these because they start with `test_`. Maybe just renaming the 3 here would fix it.

src/wrappers/OsipiBase.py
IvanARashid: Agreed. Good spotting, fixed!

etpeterson:
> > > > Can we make the tests in tests/unit_tests/test_ivim_fit.py pass? Currently 2 fail. Also, how is tests/unit_tests/simple_test_run_of_algorithm.py supposed to be used? Currently I get a fixture error if I run pytest in that file.
> > > 
> > > 
> > > I believe the two that fail is due to flaws in the algorithm. For example, if you uncomment the lines that have to do with negative values, we only get one fail. In the results, we can see that it is the ETP algorithm that fails, while the IAR alorithm doesn't. The difference between the two is that ETP does a linear fit, while IAR is a simple bi-exponential NLLS fit. So if IAR passes, then it's probably something with the linear fit that can't handle those specific testing parameters. You can try to play around with the tolerances and see if you can get ETP to pass. In the end, I think Eric knows more about the issue than I do.
> > > Regarding simple_test_run_of_algorithms, this is a simple "run" script, which initiates and performs a single fit for the algorithms implemented in OsipiBase. It was used when I wrote the standardized versions, just to see if they worked. I added it to the repo as it could be useful for doing similar test runs when we implement the rest of the submissions according to OsipiBase. So it is not supposed to be ran with pytest.
> > 
> > 
> > Regarding the failing tests, I haven't looked closely at the code yet, but I think there are a few things we need to figure out. First, I'm a little suspicious of the failures because isn't the same test data passing in `main`? Regardless of the details though, I think this is showing that not all algorithms can handle all data to the same degree and we need to be flexible about that. I think the best approach would be to have some default test values for each algorithm that we can load and use for testing. That would preserve the unit testing we want to do and also allow customized bounds (or other customizations) for each algorithm. Failing that I think we'd just need to be too permissive on the testing side to generate data and the value of the testing becomes basically nothing. I'm also OK just removing the linear fitting from the list to be tested and replacing it with another, more capable algorithm. IMO it would be interesting to include as a baseline reference algorithm, but if it's causing too much trouble then let's take the easy path for now.
> 
> I think it's good to see some fails. It means that our tolerances are working. I don't know how it affects the results though. Can we still log the estimates, etc?

All the tests should "pass". For example, if 3 are failing, who is tracking which 3 are failing? What if one starts working again, but a different one then starts to fail so we still have 3 failing? And who is tracking that it's just 3 failing? If we expect tests to fail then we should mark them as expected to fail, which in effect makes them "pass". https://docs.pytest.org/en/7.1.x/how-to/skipping.html That's a valid test, we just need to mark the ones that we expect to fail in some way.

src/wrappers/OsipiBase.py
IvanARashid: Could add that in the next iteration. I'm personally a fan of self explanatory function names, in comparison to just having str(OsipiBase).

src/wrappers/OsipiBase.py
IvanARashid: All of the checks are still very much WIP and are untested, non-paid-attention-to draft code :) I prioritized getting the testing functionality up for the abstract. But this is next on the list of things to do. If I remember correctly, no checks are being run in any of the non-check methods, unless the user calls them. Unless you see something being ran, I hope we can disregard them from this PR. I could also do one better and remove them from the current branch. What do you think?

tests/IVIMmodels/unit_tests/simple_test_run_of_algorithm.py
IvanARashid: I renamed them, but `python -m pytest` still seems to do 81 tests. Or what was it you were looking at?

src/wrappers/OsipiBase.py
etpeterson: Sounds good. I'd say keep them in because it's something we want. Maybe a halfway point would be to comment out the current implementation and always return `True`. That way there's an indication of the goal without the specific implementation details fixed.

IvanARashid: I'll do that then!

src/wrappers/OsipiBase.py
IvanARashid: I've added a check to see if the root dir is in pythonpath or not. Unfortunately I don't have the time to find a better solution right now, but will attempt to for the next version that includes proper checks. If this is the only thing holding up the PR, I think we should let it pass so we can continue the work on testing

etpeterson: I think the failing tests is the remaining thing. Easiest approach is to not use the ETP method. Perhaps I could work on it later to reintegrate it.

DKuppens:
> > > > > Can we make the tests in tests/unit_tests/test_ivim_fit.py pass? Currently 2 fail. Also, how is tests/unit_tests/simple_test_run_of_algorithm.py supposed to be used? Currently I get a fixture error if I run pytest in that file.
> > > > 
> > > > 
> > > > I believe the two that fail is due to flaws in the algorithm. For example, if you uncomment the lines that have to do with negative values, we only get one fail. In the results, we can see that it is the ETP algorithm that fails, while the IAR alorithm doesn't. The difference between the two is that ETP does a linear fit, while IAR is a simple bi-exponential NLLS fit. So if IAR passes, then it's probably something with the linear fit that can't handle those specific testing parameters. You can try to play around with the tolerances and see if you can get ETP to pass. In the end, I think Eric knows more about the issue than I do.
> > > > Regarding simple_test_run_of_algorithms, this is a simple "run" script, which initiates and performs a single fit for the algorithms implemented in OsipiBase. It was used when I wrote the standardized versions, just to see if they worked. I added it to the repo as it could be useful for doing similar test runs when we implement the rest of the submissions according to OsipiBase. So it is not supposed to be ran with pytest.
> > > 
> > > 
> > > Regarding the failing tests, I haven't looked closely at the code yet, but I think there are a few things we need to figure out. First, I'm a little suspicious of the failures because isn't the same test data passing in `main`? Regardless of the details though, I think this is showing that not all algorithms can handle all data to the same degree and we need to be flexible about that. I think the best approach would be to have some default test values for each algorithm that we can load and use for testing. That would preserve the unit testing we want to do and also allow customized bounds (or other customizations) for each algorithm. Failing that I think we'd just need to be too permissive on the testing side to generate data and the value of the testing becomes basically nothing. I'm also OK just removing the linear fitting from the list to be tested and replacing it with another, more capable algorithm. IMO it would be interesting to include as a baseline reference algorithm, but if it's causing too much trouble then let's take the easy path for now.
> > 
> > 
> > I think it's good to see some fails. It means that our tolerances are working. I don't know how it affects the results though. Can we still log the estimates, etc?
> 
> All the tests should "pass". For example, if 3 are failing, who is tracking which 3 are failing? What if one starts working again, but a different one then starts to fail so we still have 3 failing? And who is tracking that it's just 3 failing? If we expect tests to fail then we should mark them as expected to fail, which in effect makes them "pass". https://docs.pytest.org/en/7.1.x/how-to/skipping.html That's a valid test, we just need to mark the ones that we expect to fail in some way.

Nice bit of background info Eric, was not really aware of this expected to fail functionality :).

etpeterson:
Looks like it has conflicts. Otherwise good!

etpeterson:
Looks like it's all ready to merge. Nice work!

etpeterson:
Looks good! We should get this merged so the rest of the testing can be more easily based off of it.

PR #13, opened 2023-09-22
Added ADC and IVIM functions from NKI

KoenBaas:
Created folder PvH_KB_NKI. Adding DWI_Examples.py, DWI_functions_standalone.py, and IVIM_b0_15_150_500.npy.


src/original/PvH_KB_NKI/DWI_functions_standalone.py
```diff
@@ -0,0 +1,239 @@
+f# -*- coding: utf-8 -*-
```
etpeterson, line 1: This `f` seems to be causing the test parser to choke.

KoenBaas: Sorry for the delay. I removed the `f` that was causing this error.

etpeterson: Perhaps just change the path to the current directory so the function runs. Maybe add a comment with the YOUR_PATH so people know. I'll also try to see if there's a way to avoid these failures in the future.

oliverchampion:
Hey Koen, your code is now "failing" tests on Paulien's code. As a result, I could not find easiLy whether your code causes (additional) errors. I think you can remove the file "src/standardized/PV_MUMC_biexp.py" in your repository that would get rid of most of the fails.

PR #15, opened 2023-10-19
Ogc amsterdam umc standardized

oliverchampion:
added to wrapper, incluidng Bayesian fit.


IvanARashid:
What are the changes to the XCAT phantom? Do we want them?

PR #20, opened 2023-10-23
Testing/add etp and add configuration

etpeterson:


oliverchampion:
Hey! Several questions:

What is the reason for now initializing the code for every voxel?:
Adding 
self.initialize(bounds, initial_guess, fitS0, thresholds, prior_in)
to the ivim_fit routine, instead of running it in the super?
I imagine typically you want to fit a full image with similar settings. Then, if you want to change bounds, you could run initialize again after fitting the image?

oliverchampion:
Moreover, the Bayesian approach needs an initial guess that is reasonable. This is now taken from the segmented fit. If the segmented fit is broken, it is expected that the Bayesian will not work well. :)

oliverchampion:
I think for the Bayesian wrapper you can remove line 77 "        fit_results=fit_results+(1,)"

oliverchampion:
Unfortunately, I cannot run the tests on segmented fit without the json file (which is now being ignored by .gitignore).
My expectation is that the bounds implemented for my segmented fit are not working well (i.e. not well translated to the individual fits). Does the segmented fit work for the fit bounds given in my wrapper?

oliverchampion:
I think I've fixed the errors in segmented and bayesian. Still confused about the initialisation.

PR #21, opened 2023-10-26
GU code contribution

oscarjalnefjord:
GU code contribution with segmented fitting. I had to make a small adjustment of the test code such that it tests each parameter separately. Maybe not a problem with the new test code. Let me know otherwise.


etpeterson:
> GU code contribution with segmented fitting. I had to make a small adjustment of the test code such that it tests each parameter separately. Maybe not a problem with the new test code. Let me know otherwise.

That sounds strange, but I also did that in my test to apply different thresholds to each. I haven't looked at it thoroughly, but tests pass so I guess that's good for now.

I'd say merge.

PR #27, opened 2023-11-06
GU code contributions

oscarjalnefjord:
Including the previously missing algorithm (with wrapper adjustment to not cause test fail to due wrong shape)


etpeterson:
Looks good to me. Most likely the analysis will work as well.

PR #31, opened 2024-01-19
>1D input support

IvanARashid:
Added support for >1D inputs to the wrapper. Would be appropriate if it was tested by actually fitting an image.

It is quite slow though. I'm getting 300 fits/s. But that's probably because of all the ways we are calling functions through the wrapper code?

### DIFF
--- a/src/wrappers/OsipiBase.py
+++ b/src/wrappers/OsipiBase.py
@@ -47,7 +47,7 @@ class OsipiBase:
         pass
 
     #def osipi_fit(self, data=None, bvalues=None, thresholds=None, bounds=None, initial_guess=None, **kwargs):
-    def osipi_fit(self, data, bvalues, **kwargs):
+    def osipi_fit(self, data, bvalues=None, **kwargs):
         """Fits the data with the bvalues
         Returns [S0, f, Dstar, D]
         """
@@ -68,7 +68,6 @@ class OsipiBase:
         #kwargs["bvalues"] = use_bvalues
         
         #args = [data, use_bvalues, use_thresholds]
-        args = [data, use_bvalues]
         #if self.required_bounds or self.required_bounds_optional:
             #args.append(use_bounds)
         #if self.required_initial_guess or self.required_initial_guess_optional:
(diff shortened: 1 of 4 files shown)


tests/IVIMmodels/unit_tests/simple_test_run_of_algorithm.py
```diff
@@ -1,11 +1,17 @@
-import numpy as np
+import sys
+sys.path.append(r"C:\Users\ivan5\Box\OSIPI\TF2.4_IVIM-MRI_CodeCollection")
```
etpeterson, line 2: This looks like a fixed path on your local machine.

etpeterson:
> Added support for >1D inputs to the wrapper. Would be appropriate if it was tested by actually fitting an image.
> 
> It is quite slow though. I'm getting 300 fits/s. But that's probably because of all the ways we are calling functions through the wrapper code?

Regarding testing, what about a small multi-dimensional test? Not a whole image, but a 10x10 synthetic patch?

And regarding speed, what algorithms? In the current testing pipeline we record speeds and I see your algorithms going at about that pace. Could it help to cache the inputs in some way? It just makes me think of this decorator, but I can't think of how to use it at the moment. https://docs.python.org/dev/library/functools.html#functools.lru_cache

src/wrappers/OsipiBase.py
```diff
@@ -83,7 +83,13 @@ def osipi_fit(self, data, bvalues, **kwargs):
         
         #args = [data, use_bvalues, use_initial_guess, use_bounds, use_thresholds]
         #args = [arg for arg in args if arg is not None]
-        results = self.ivim_fit(*args, **kwargs)
+
+        # Assuming the last dimension of the data is the signal values of each b-value
+        results = np.empty(list(data.shape[:-1])+[3]) # Create an array with the voxel dimensions + the ones required for the fit
+        for ijk in tqdm(np.ndindex(data.shape[:-1]), total=np.prod(data.shape[:-1])):
```
etpeterson, line 89: I'm not sure if we have algorithms that accept 2+D inputs, but at some point we'll need to iterate on the correct number of dimensions here. E.g. if an algorithm is 2D, this loop should only operate on the final dimension for that algorithm.

PR #32, opened 2024-01-24
Infrastructure/reduce repository size

etpeterson:
This converts some files to LFS, which will end up getting overwritten anyway. It doesn't remove them from the history, that needs to come later, but it does fix the testing pipeline for use with LFS files.


etpeterson:
Actually, don't merge yet. I think we shouldn't be pulling lfs in testing. Will need to fix that at the source.

PR #38, opened 2024-02-26
Updated testing b-values with a more realistic set

### LINKED ISSUE #10: Update testing b-values with a more realistic set
They're currently skewed high, so should be updated with literature supported values.

AhmedBasem20:
Fixes #10
Tried to update the unit testing b-values with literature-supported values. the values are obtained from a public dataset.
I found other values as well in this paper: https://onlinelibrary.wiley.com/doi/10.1002/mrm.26783 

>  12 b-values (1.4, 5, 10, 20, 35, 50, 75, 100, 201, 401, 602, and 802 s/mm2) 

@etpeterson  @oliverchampion  Not sure which one has the better representation so please try both and let me know.

### Values source:
https://figshare.com/articles/dataset/IVIM_dataset/3395704?file=5305246


etpeterson:
Unfortunately that whole file needs to be regenerated if the b-values change. I responded in more detail in the [issue](https://github.com/OSIPI/TF2.4_IVIM-MRI_CodeCollection/issues/10).

PR #51, opened 2024-03-04
Generate a variety of testing data and came up command_line version of sim_vim_sig.py

### LINKED ISSUE #41: Generate a variety of testing data
### Feature description

Currently the only testing data we have is a single b-value set, we want a variety to choose from.

### Describe the solution

We generated a csv testing file from binary files now on [zenodo](https://zenodo.org/records/10696605) using [this](https://github.com/OSIPI/TF2.4_IVIM-MRI_CodeCollection/blob/main/phantoms/MR_XCAT_qMRI/sim_ivim_sig.py) script. It would be good to have a variety of files to choose from. This would involve running the script with a variety of b-values, generating files, and adding them to the repository.

### Describe alternatives

_No response_

### Additional context

_No response_

### Are you working on this?

None

Unique-Usman:
Added this two files. 
sim_ivim_sig_command_line.py - This is the command line version of the sim_vim_sig,py
sim_ivim_sig_mul_bval.py - This version is used for running multiple b_value which are stored in b_values.py
                                                                   
Various test files based on the multiple value of b_value in the b_values.py
├── generic_five.json
├── generic_four.json
├── generic_original.json
├── generic_three.json
├── generic_two.json

Also, the generic_original.json was supposed to be the same thing as generic.json. But, the result is not the same. I  guess the generic.json was generate from another b_value different from the current one. 

Also, @etpeterson what do you think about combining the three scripts(sim_ivim_sig_command_line.py, sim_ivim_sig_mul_bval.py, sim_ivim_sig.py) together instead of having multiple scirpts doing almost similar thing ? What we can do is to just have a single script that can be used as command line tools and is able to generate multiple value of test data based on the multiple b_values. 

I will update the pr based on your feeedback and also come up with a documentation for running the scipt. 

Fixes #41

### Checklist

- [ ] Update Docs & Guides


etpeterson:
Yes, I think it should just be one script, just changes to the original `sim_ivim_sig.py` would be great. One of the inputs could be a choice for bvalues to generate.

The change to `generic.json` looks like somehow the newline at the end got dropped, but the contents are the same. It's hard to say how this happened because it's difficult to see changes with the duplicated script. Once everything is all together then it'll be easier to look at the changes made.

phantoms/MR_XCAT_qMRI/sim_ivim_sig.py
```diff
@@ -103,7 +105,7 @@ def contrast_curve_calc():
     D[36] = 3e-3  # 36 artery
     D[37] = 3e-3  # 37 vein
     D[40] = 1.31e-3  # 40 asc lower intestine : Hai-Jing et al. doi: 10.1097/RCT.0000000000000926
-    D[41] = 1.31e-3  # 41 trans lower intestine : Hai-Jing et al. doi: 10.1097/RCT.0000000000000926
+    D[41] = 1.31e-3  # 41 trans lower# If conversion fails, try splitting the input as a lis intestine : Hai-Jing et al. doi: 10.1097/RCT.0000000000000926
```
etpeterson, line 108: Probably accidental add?

phantoms/MR_XCAT_qMRI/sim_ivim_sig.py
```diff
@@ -364,76 +366,125 @@ def XCAT_to_MR_DCE(XCAT, TR, TE, bvalue, D, f, Ds, b0=3, ivim_cont = True):
...
-    # Save the NIfTI image to a file
-    nifti_img.header.set_data_dtype(np.float64)
-    if not motion:
-        output_file = 'output.nii.gz'  # Replace with your desired output file name
-    elif interleaved:
-        output_file = 'output_resp_int.nii.gz'  # Replace with your desired output file name
+    parser.add_argument("-b", "--bvalue", type=float,
+                        nargs="+",
+                        help="B values (list of of numbers)")
+    parser.add_argument("-bf", "--bvalues-file", metavar="FILE", type=parse_bvalues_file,
```
etpeterson, line 398: Maybe just `-f` for the short?

phantoms/MR_XCAT_qMRI/b_values.json
```diff
@@ -0,0 +1,7 @@
+[
+    [0.0, 1.0, 2.0, 5.0, 10.0, 20.0, 30.0, 50.0, 75.0, 100.0, 150.0, 250.0, 350.0, 400.0, 550.0, 700.0, 850.0, 1000.0],
```
etpeterson, line 2: Rather than a list of lists perhaps an object so they can have names rather than just numbers?

tests/IVIMmodels/unit_tests/generic.json
etpeterson, line 1: I don't understand your comment that the file changed. It looks like just a newline to me, am I wrong?

Along those lines, whichever file is a duplicate of this should be deduplicated. Your changes shouldn't alter the original bvalues in any way.

phantoms/MR_XCAT_qMRI/sim_ivim_sig.py
```diff
@@ -364,76 +366,125 @@ def XCAT_to_MR_DCE(XCAT, TR, TE, bvalue, D, f, Ds, b0=3, ivim_cont = True):
...
+        raise argparse.ArgumentError(None, "Arguments --bvalues-file and --bvalues are mutually exclusive")
+    
+    bvalues = None
+    if args.bvalues_file:
+        bvalues = args.bvalues_file
+    elif args.bvalue:
+        bvalues = [args.bvalue]
     else:
-        output_file = 'output_resp.nii.gz'  # Replace with your desired output file name
+        bvalues = [[0., 1, 2, 5, 10, 20, 30, 50, 75, 100, 150, 250, 350, 400, 550, 700, 850, 1000]]
```
etpeterson, line 414: It seems odd to have a double array, why is this?

tests/IVIMmodels/unit_tests/generic_0.json
```diff
@@ -0,0 +1,251 @@
...
+        "Dp": 0.009999999999999998,
+        "data": [
+            0.10069801124430204,
+            0.11354496636645572,
+            0.11515572103570605
+        ]
+    },
+    "config": {
+        "bvalues": [
+            43.0,
```
etpeterson, line 246: Something isn't right with these bvalues.

etpeterson (changes requested):
Getting there. Several specific comments and it looks like one of the json files isn't correctly generated.

phantoms/MR_XCAT_qMRI/sim_ivim_sig.py
Unique-Usman: Yeah, it is accidental. Will correct it.

phantoms/MR_XCAT_qMRI/sim_ivim_sig.py
Unique-Usman: Yeah, noted. Will make changes to it.

phantoms/MR_XCAT_qMRI/b_values.json
Unique-Usman: I tried object for the naming, but, we will have to enforce a pattern of naming on any one running the command for easy parsing. 

Something like { generic_<name> : [], ...}. That is why I made it into a list of list. But, if you think object is better approach then I can change it to that.

phantoms/MR_XCAT_qMRI/sim_ivim_sig.py
Unique-Usman: I had to make it a double array so that I can have one implementation with when the value is gotten from the files. Check line 420.

tests/IVIMmodels/unit_tests/generic_0.json
Unique-Usman: Ohh, that is true, I will change it. Those values I gave while testing the script in the command line.

tests/IVIMmodels/unit_tests/generic.json
Unique-Usman: The generic.json does not change. What I am saying is that, when I was trying to regenerate the generic.json using the current bvalues. I got different json values. So, that is why I said the generic.json might be generated from another bvalues different from the placeholders.

phantoms/MR_XCAT_qMRI/b_values.json
etpeterson: I'd prefer objects. That way we have meaningful file names, like generic_short.json.

Some naming system would be good. I'm not sure what you mean "for easy parsing". I would imagine that people give either no name and they get all out or they give a name and they get that one out.

phantoms/MR_XCAT_qMRI/sim_ivim_sig.py
etpeterson: Oh I see. I guess it'll change with an object implementation anyway.

etpeterson:
> What I was saying regarding the generic.json is that, the value in generic_0.json was supposed to be the same thing with the generic.json. Because, the bvalue for generic_0.json was the one bvalue in the repo. My suggestion is that, the bvalue used to generate generic.json is different from the one present in the repo.

Is this still different? We can see the bvalues used for generic.json in the [file itself](https://github.com/OSIPI/TF2.4_IVIM-MRI_CodeCollection/blob/main/tests/IVIMmodels/unit_tests/generic.json#L575). And I can see they are the same as the default set you're using. Is it different on the main branch too? I would be surprised if it's different on the main branch so that would be a good place to check.

tests/IVIMmodels/unit_tests/generic.json
etpeterson: There shouldn't be changes to generic.json. I commented elsewhere on that. If it also changed in the main branch then perhaps, but I would be very surprised if that were the case.

Assuming there aren't changes though, we don't need two copies of generic.json. So confirming a duplicate and remove the original `generic.json` and point everything to the new one.

Unique-Usman:
@etpeterson , I ran on it on the main branch and it still give the same result.

etpeterson:
> @etpeterson , I ran it on the unchaged main branch and it still give the different result from the generic.json. Yeah, the bvalues used is the same, something is wrong somewhere I guess.

That is surprising. What's the difference?

Unique-Usman:
All the decimal values
Screenshot from 2024-03-07 11-32-45

etpeterson:
This is strange. I don't know why that would change. The history doesn't show meaningful changes to that file. I'll give it a shot.

Also, unfortunately you have a merge conflict.

etpeterson:
I'm not sure what to say, but it worked for me, I got the same data in both. Here are my steps.

- from the `phantoms/MR_XCAT_qMRI` folder
- `git checkout main`
- `git pull`
- `python sim_ivim_sig.py`
- let it run and in a minute or two it finished
- `git status` shows no changes
- `diff generic.json ../../tests/IVIMmodels/unit_tests/generic.json` shows no changes

Unique-Usman:
Did you move the utitlities to the `phantoms/MR_XCAT_qMRI` when running the code ?

etpeterson:
> Did you move the utitlities to the `phantoms/MR_XCAT_qMRI` when running the code ?

No, if the pythonpath is set correctly it finds them where they are. Even if you pull the images manually I think the result should be the same though.

Unique-Usman:
Yeah, I was able to get the same data. But, do we want  to add https://github.com/OSIPI/TF2.4_IVIM-MRI_CodeCollection/blob/main/phantoms/MR_XCAT_qMRI/sim_ivim_sig.py#L13 to the script. This keeps downloading all the data everytime.

etpeterson:
> This keeps downloading all the data everytime.

If that's the case then it sounds like a good issue to be fixed.

etpeterson:
I'll try to actually run it later and test that, but the code looks good to me at the moment.

etpeterson:
> Yeah, I was able to get the same data

What was the problem? Is it something that we can fix with some instructions or code?

tests/IVIMmodels/unit_tests/generic.json
etpeterson: This is still pending.

phantoms/MR_XCAT_qMRI/b_values.json
etpeterson: This is still pending as well.

Unique-Usman:
> f that's the case then it sounds like a good issue to be fixed in a different pull request.

I really think it is some mix up somewhere, probably I deleted or add some code mistakenly. I resolved the problem by rewriting the code from the start.

phantoms/MR_XCAT_qMRI/b_values.json
Unique-Usman: Ok, I will change it to object.

etpeterson:
> I think, if possible, we should have the downloading of data separately. Then, in the documentation for running the code that I will be providing, I will mention it there.

We shouldn't try to reimplement git-lfs here, but also a simple check to see if the requisite files are present and if not then download seems pretty easy. If it's in the document and in the code then requiring two steps is fine.

PR #55, opened 2024-03-07
Generate a variety of testing data and came up command_line version of sim_vim_sig.py

### LINKED ISSUE #10: Update testing b-values with a more realistic set

Unique-Usman:

Update to this pull request - https://github.com/OSIPI/TF2.4_IVIM-MRI_CodeCollection/issues/41

Fixes #10


etpeterson:
Is this a duplicate of your other one? https://github.com/OSIPI/TF2.4_IVIM-MRI_CodeCollection/pull/51

Unique-Usman:
> Is this a duplicate of your other one? #51

No, This is the correct branch that solve the issue. I had to start from fresh. I can close the other one.

etpeterson (changes requested):
The changes still pending on the other PR, then I think we're good. Those being objects rather than indices and deduplicating generic.json from generic_0.json or whatever it duplicate is.

etpeterson:
> I can close the other one.

Sounds good.

etpeterson:
I see you made changes but because you force pushed and squashed them into a single commit I can't see your changes. In general try to avoid force pushes and history rewriting. I also commented about that to someone else and you can see my reasoning [here](https://github.com/OSIPI/TF2.4_IVIM-MRI_CodeCollection/pull/53#issuecomment-1984015769).

phantoms/MR_XCAT_qMRI/sim_ivim_sig.py
```diff
@@ -366,76 +368,123 @@ def XCAT_to_MR_DCE(XCAT, TR, TE, bvalue, D, f, Ds, b0=3, ivim_cont = True):
...
+                        raise argparse.ArgumentTypeError("bvalues in JSON file are not list")
+                    for value in bvalue:
+                        if not isinstance(value, float):
+                            raise argparse.ArgumentTypeError("Values in lists are not float")
+        except json.JSONDecodeError as e:
+            raise argparse.ArgumentTypeError(f"Invalid JSON file: {e}")
+
+        return bvalues_dict
 
+    parser.add_argument("-b", "--bvalue", type=float,
```
etpeterson, line 396: Perhaps default to the file b_values.json and the "original" set. Then the if/else below can also be removed.

etpeterson (changes requested):
Looking good, I was able to run it on my machine and it works. Just the default change and I think that's it.

Unique-Usman:
> Contributor

I am very sorry, I will take note of that. Your reason make sense. Having all the commits is also a good way to follow the commit process for a branch/changes. Thanks for the knowledge.

phantoms/MR_XCAT_qMRI/sim_ivim_sig.py
Unique-Usman: I think there wil still be if/else. It will look something like this.

`
 bvalues = None
    if args.bvalues_file:
        bvalues = args.bvalues_file
    elif args.bvalue:
        bvalues = {"cmd": args.bvalue}
    else:
       bvalues = parse_bvalues_file("b_values.json")
`

If we have default for both of the value(-f, -b), they will both give error due to :- 

`
    if args.bvalues_file and args.bvalue:
        raise argparse.ArgumentError(None, "Arguments --bvalues-file and --bvalues are mutually exclusive")
`

etpeterson: Ah, you're right. Maybe just default to using the "original" set from the file then? It seems the most standard.

PR #56, opened 2024-03-08
report on repo

### LINKED ISSUE #47: Automatically report on repository status
### Feature description

A simple report on the repository containing information summarizing the repository like how many submissions it contains and how many are being tested could be useful. It could also contain a little bit of information about each submission, if that is provided.

### Describe the solution

A simple html or pdf report that summarizes the repository statistics would be nice to have. The number of submissions, number of wrapped (tested) submissions, and a little bit of information about each would be a good start.

This could initially just be generated as a workflow artifact, but could also be pushed to a github page as well.

I imagine this as being a small python "test" that parses the repository for this information and generates a test artifact. It could be run with each merge to the main branch to update the information every time the repository changes.

### Describe alternatives

_No response_

### Additional context

_No response_

### Are you working on this?

None

abhicodes369:
created a report on the repository like the name of the contributor and how many are being tested 
a simple HTML is being generated if you execute this script 

Fixes #47 

### Checklist

- [ ] Update Docs & Guides


combined_report.html
etpeterson, line 1: This should go into the `website` folder. Another PR is also creating files there.

local.env.example
etpeterson, line 1: What is the idea behind setting these variables? You should be able to get the repository location from the location of the file, e.g. using the `__file__` variable.

repostatus.py
etpeterson, line 1: This should move somewhere else. Probably `utilities/repository` or something.

repostatus.py
```diff
@@ -0,0 +1,37 @@
+import os
+import pandas as pd
+from dotenv import load_dotenv
+import json
+
+# Load environment variables
+load_dotenv(r'C:\Users\home\tf2.4\TF2.4_IVIM-MRI_CodeCollection\local.env')
```
etpeterson, line 7: We should be able to get around hard coding paths. This won't work for me, for example.

repostatus.py
```diff
@@ -0,0 +1,37 @@
...
+
+# Read the CSV file
+csv_file_path = os.getenv('CODE_CONTRIBUTIONS_FILE')
```
etpeterson, line 10: Looking through the code I realize that your table is missing a column, `wrapped`. In the [issue thread](https://github.com/OSIPI/TF2.4_IVIM-MRI_CodeCollection/issues/47#issuecomment-1973909276) I mentioned parsing the files in the folders which isn't being done here, and that's what is missing. It should parse the wrapper folder to find which algorithms are wrapped.

And there I also mentioned double checking the code contributions against what is in the source folder.

So the source code folder and code contribution file should match. If not it would be nice to note that. That should be a superset of the wrapped code list. And that should be a superset of the tested code list. If anything doesn't match that logic then we should note it in the html or printing a message.

etpeterson (changes requested):

repostatus.py
```diff
@@ -0,0 +1,37 @@
+import os
+import pandas as pd
```
etpeterson, line 2: Pandas caused the test failure. It should be in the requirements.

abhicodes369:
I apologize I accidentally deleted the git ignore file can I create a new branch

etpeterson:
> I apologize I accidentally deleted the git ignore file can I create a new branch

Sure, start over or just add it back on this branch, either one.

utilities/repostatus.py
```diff
@@ -0,0 +1,64 @@
...
+all_algorithms = algorithms_data['algorithms']
+
+# Check if both code_contributions_file matches with source folder
+for subfolder in unique_subfolders:
+    subfolder_path = os.path.join(SOURCE_FOLDER, subfolder)
+    if not os.path.exists(subfolder_path):
+        print(f"Warning: Subfolder '{subfolder}' does not exist in the source folder.")
+
+# Add column 'Tested' to the DataFrame if it starts with that of subfolder
+df['Tested'] = df.apply(lambda row: 'Yes' if any(algorithm.startswith(row['subfolder'].split('_')[0]) for algorithm in all_algorithms) else 'No', axis=1)
```
etpeterson, line 35: I'm just looking at this logic and it seems like it's just matching everything before the first `_`. That's not specific because that's usually the author's initials and they can have several algorithms.

Looking at it a little more, this could become fancier, but maybe we actually want this column in our csv file. So another approach would be to just manually go through and find the matches and add a "Wrapped" column to the csv file that contains what you'd see in the algorithms.json.

Then that would reduce what is needed in here to just having a sanity check that the wrapped folders match the csv. And it would make checking for tested much easier because it would be a simple string match.

How does that sound?

utilities/repostatus.py
```diff
@@ -0,0 +1,64 @@
...
+
+# Parse files in the WRAPPED_FOLDER 
+wrapped_algorithms = []
+for root, dirs, files in os.walk(WRAPPED_FOLDER):
```
etpeterson, line 39: This could just be added to the csv, see comment above.

utilities/repostatus.py
```diff
@@ -0,0 +1,64 @@
...
+        if file.endswith('.py'):
+            file_path = os.path.join(root, file)
+            with open(file_path, 'r') as f:
+                content = f.read()
+                for algorithm in all_algorithms:
+                    if algorithm in content:
+                        wrapped_algorithms.append(algorithm)
+
+# Add a column 'Wrapped' to the DataFrame
+df['Wrapped'] = df.apply(lambda row: 'Yes' if any(algorithm.startswith(row['subfolder'].split('_')[0]) for algorithm in wrapped_algorithms) else 'No', axis=1)
```
etpeterson, line 50: This as well, see above.

utilities/repostatus.py
```diff
@@ -0,0 +1,64 @@
...
+
+# Select the desired columns
+df_selected = df[['Technique', 'subfolder', 'Authors', 'Tested', 'Wrapped']]
+df_selected.columns = ['Technique', 'Subfolder', 'Contributors', 'Tested', 'Wrapped']
+
+# Convert the DataFrame to HTML
+html_string = df_selected.to_html(index=False)
+
+# Save the HTML to a file
+with open(os.path.join(REPO_DIR, 'combined_report.html'), 'w') as f:
```
etpeterson, line 60: I think `os.path.join(REPO_DIR, website, 'combined_report.html')`

etpeterson (changes requested):
Sorry, I'm rethinking this a little. Some of these that you're trying to find could be useful in the csv file anyway, so perhaps just adding them could simplify the script a lot. Then it could be a simpler process to generate the html from the csv.

utilities/repostatus.py
```diff
@@ -0,0 +1,64 @@
+import os
+import pandas as pd
+import json
+
+# directory of the current script
+script_dir = os.path.dirname(os.path.abspath(__file__))
```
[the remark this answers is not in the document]
abhicodes369: Would it be problematic to add a column called 'algorithms' to the data frame? it would be easily understandable. Is it a bad idea?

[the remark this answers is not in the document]
etpeterson: What would that add? The algorithms from the json are the ones running the automated testing. That's beyond the scope of the csv file. Is there some other algorithm you want to add?

utilities/repostatus.py
abhicodes369: do i need to manually update the code_contribution_csv file or write a script to read algorithms from algorithms.json and write 'wrapped' in csv file   and automatically update whenever there is change in algorithms. json file

etpeterson: Add the column manually. That column is not the same as what's in algorithm.json. That column would be the wrapped name, so if an algorithm is wrapped it should have an entry there. The algorithm.json is what is being tested. In theory everything wrapped should be tested but it does need to be added and confirmed it doesn't fail. So they're not the same thing.

etpeterson:
This is looking good, I think you're almost there. After these small changes I'll send it around to the contributors for a final sanity check.

etpeterson:
Just checking on this, are these changes still in the works?

abhicodes369:
image
it looks like this.
Are there any modifications required?

etpeterson:
> image
> it looks like this.
> Are there any modifications required?

I had 3 comments that look like they haven't been addressed yet. Small but useful like choosing empty rather than no.

abhicodes369:
i am sorry but i couldnt find these comments could you please give the link to comments that state necessary changes

etpeterson (changes requested):
Sorry! It was my fault, I didn't finalize the review.

etpeterson:
> i am sorry but i couldnt find these comments could you please give the link to comments that state necessary changes
> 

Apologies, I didn't finish the review so my comments didn't show for you.

etpeterson:
Looks like a merge conflict in `requirements.txt`.

utilities/repostatus.py
```diff
@@ -0,0 +1,50 @@
+import os
```
etpeterson, line 1: 2 things for this file.
1. Could you wrap the work up into a function so testing doesn't run it each time? Something like this:
```
if __name__ == "__main__":
    generate_html()
```
2. Could you add some documentation about how to run it and what it does.

etpeterson (changes requested):
Looks good, just those changes then I'll run it locally as a final test.

etpeterson (approved):

PR #49, opened 2024-03-02
Display figures on a new github pages site.

AhmedBasem20:
### Describe the changes you have made in this PR

Displayed the generated plots from the analysis workflow to a new github-pages site.
### Results
Example from my fork: https://ahmedbasem20.github.io/TF2.4_IVIM-MRI_CodeCollection/

@etpeterson To enable this on this repository we need the following steps:

1. create a new branch named `gh-pages`.
2. From the pages section on the repository settings, set `gh-pages` as a source branch.
image

### Link this PR to an issue [optional]

Fixes #46

PR #59, opened 2024-03-13
Generate API documentation using Sphinx.

### LINKED ISSUE #50: Documentation generation
### Feature description

There's currently no documentation being generated. It would be nice to have some basics at least.

### Describe the solution

This would involve using pydoc or sphinx to build documentation. It might be tricky because the original source may have no documentation, or documentation of differing types. This might need to be ignored. The wrapper code might have some useful documentation, but may also need expanding before it is truly very useful.

### Describe alternatives

_No response_

### Additional context

This may be more difficult than expected because of the varieties in coding styles and the wrapping. It might also require more baseline documentation to be written before it is useful.

### Are you working on this?

None

AhmedBasem20:

Implemented automated API documentation generation using Sphinx and Github Actions. This process involves two steps:

1. Executing `sphinx-apidoc -o doc src` to generate initial API documentation into `doc` folder after extracting docs from `src` folder.
2. Running `make html` to generate HTML documentation in the `doc/_build/html` folder.

This setup allows for including manual documentation alongside API documentation. I'll try to integrate the markdown files from the `doc` folder onto the documentation page. And exclude the `src/original` file as it doesn't have much documentation there.
### Documentation Link
https://ahmedbasem20.github.io/TF2.4_IVIM-MRI_CodeCollection

Fixes #50


.github/workflows/sphinx.yml
```diff
@@ -0,0 +1,51 @@
...
+        if: steps.pip-cache.outputs.cache-hit != 'true'
+
+      - name: Install dependencies
+        run: |
+          pip install -r requirements.txt
+
+      - name: Build html
+        run: |
+          sphinx-apidoc -o doc src
+          cd doc/
```
etpeterson, line 41: The `doc` folder is not empty. Maybe best to create an empty folder to use for the sphinx docs?

.github/workflows/sphinx.yml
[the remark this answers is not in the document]
AhmedBasem20: Not sure if this is possible, will test it. But It is common to deploy documentation on [readthedocs](https://readthedocs.org/) so I may give it a try.

.github/workflows/sphinx.yml
AhmedBasem20: I used it on the existing folder in case we want to integrate the existing files in `doc` into the website, but if this is not the case I'll create another folder for sphinx.

.github/workflows/sphinx.yml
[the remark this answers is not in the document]
etpeterson: I could imagine a main page that has links to the analysis and to documentation. Deploying could work too but it looks like the better case is to link accounts, and we'd probably want an official OSIPI or TF2.4 account for readthedocs in that case. https://sphinx-rtd-tutorial.readthedocs.io/en/latest/read-the-docs.html

.github/workflows/sphinx.yml
etpeterson: Mostly that there are two build steps and it's putting intermediate files into the doc folder as well as the final files? Maybe there's no real intermediate, I can't check at the moment.

But it looks like those other docs are more about git contribution and not usage. I'm thinking is best to keep sources of documentation separate and published separately.

.github/workflows/sphinx.yml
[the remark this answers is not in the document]
AhmedBasem20: > I could imagine a main page that has links to the analysis and to documentation.

Exactly, I'm trying to include the analysis in the documentation as a separate section.
image

AhmedBasem20:
Hi @etpeterson, please check the website again and let me know what you think: https://ahmedbasem20.github.io/TF2.4_IVIM-MRI_CodeCollection
We now have a main page that points to the algorithm analysis figures and the API documentation.

I used this action https://github.com/dawidd6/action-download-artifact to get the `Figures` artifact uploaded by the `analysis.yml` workflow. And then included it in the docs.
This solution supersedes #49 .

.github/workflows/sphinx.yml
```diff
@@ -35,16 +35,26 @@ jobs:
         run: |
           pip install -r requirements.txt
 
+      # Action to download artifacts from a different workflow (analysis.yml)
+      - name: Download figures artifact
```
etpeterson, line 39: This introduces a dependency between workflows that we now need to manage. Is it not possible to upload documentation into one folder on the published page from one workflow and figures from another? Or does it purge the page when a new thing is uploaded?

If we do need to merge the results from multiple workflows to publish, then that publish step should probably be a separate workflow triggered when one or both of the source workflows complete. I looks like that's possible with [this](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) and it even allows passing artifacts around, so perhaps that's how it's intended?

Another option would be to add this doc building to the analysis workflow. I'm not too happy with that because it seems like it's two separate things, but maybe that's the publishing workflow?

AhmedBasem20: > Is it not possible to upload documentation into one folder on the published page from one workflow and figures from another?

Not sure if I understand this part, can you elaborate more? The `sphinx.yml` workflow downloads the figures from the last completed `analysis` workflow. And then publish the website. So this doesn't require the current `analysis` job to be completed, it captures the figures from the last successful one.

> I looks like that's possible with [this](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) and it even allows passing artifacts around, so perhaps that's how it's intended?

This will be very useful to keep the website updated with the current figures produced by `analysis.yml`.
I'm thinking of deploying the website twice: a quick one that takes ~50 seconds and uses the last available figures artifact. And the second one is when the new figures are uploaded. What do you think?
```
on:
  push:
    branches:
      - 'main'
  workflow_run:
    workflows: ["analysis.yml"]
    types:
      - completed
```

etpeterson: This is really good. It would be nice if the upload were a separate workflow, but maybe that's a followup improvement.

What happens in the case of a build failure on either branch?

I think the last thing is it would be nice to use the official github [method](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#using-data-from-the-triggering-workflow) of accessing data from the other workflows. It's much more verbose but is the official method. Let me know your thoughts.

AhmedBasem20: > This is really good. It would be nice if the upload were a separate workflow, but maybe that's a followup improvement.

I did the build and deployment on the same workflow because each step depends on the previous one. Not sure how to further improve this, maybe upload the build folder as an artifact and use it on a separate deployment workflow?


>  What happens in the case of a build failure on either branch?

If the build fails for any reason the website will be available with the last successful build.

> I think the last thing is it would be nice to use the official github [method](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#using-data-from-the-triggering-workflow) of accessing data from the other workflows. It's much more verbose but is the official method. Let me know your thoughts.

Done!

etpeterson (changes requested):
Looking good, just those small comments.

.github/workflows/analysis.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - 'main'
       - 'analysis/**'
+      - 'test-docs'
```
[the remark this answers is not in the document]
AhmedBasem20: Oh sorry, this was a branch that I was debugging in. Removed.

.github/workflows/sphinx.yml
AhmedBasem20: Fixed! Now if the analysis fails, the figures step will be skipped and the generation will work.

etpeterson:
> @etpeterson this pull request should replace #49, as we already generate the figures here but inside the sphinx theme.

Ah, I see. Ok, close the other one. Let's give a few more days to suggest locations and we can merge this.

PR #66, opened 2024-06-01
created the dockerise version of the TF2.4_IVIM-MRI_CodeCollection

### LINKED ISSUE #64: [Feature] <Docker container>
### Feature description

Dockerize the code so that processing could be run from the command line, e.g. docker run <TF2.4image> <IVIM input images> <IVIM output images>

### Describe the solution

This would simply put the OSIPI code inside of a docker container that could be run on the command line. Something like this for now: https://stackoverflow.com/questions/42625947/docker-input-output-outside-the-container

Just to run one and exit. Future work would be to keep it alive for future connections.

### Describe alternatives

The current approach doesn't use the docker container. This would allow for someone to run the processing without installing anything other than docker and downloading our image. Further work could make it more of a server architecture, but this is the first step.

Other containerization could be used as well, such as singularity. Perhaps both could be supported in the future.

### Additional context

_No response_

### Are you working on this?

None

Unique-Usman:
###Created the dockerise version of the TF2.4_IVIM-MRI_CodeCollection
Fixes #64

@etpeterson Kindly check the Dockerfile for any changes and also the the .dockerignore for files you think we can ignore.

When I build this, the size of the whole image was about 6.3gb which I think might be too big. 

As of now it works perfectly.

Also, which repository are we pushing the image to ? 

Also, I remove the loggin in the requirement.txt as it is a builtin python library and it is causing issue. 

Thank you.


requirements.txt
```diff
@@ -2,7 +2,6 @@ numpy
 scipy
 torchio
 torch
-logging
```
etpeterson, line 5: What was the problem here? I'm not sure where it's used but it also seems strange that it would cause issues.

Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM python:3.9-slim
```
etpeterson, line 1: Maybe Python 3.11 just because it's the newest stable version? Or a variable with that as a default.

.dockerignore
etpeterson, line 1: I'd move these files into their own `Docker` folder. You can still build from the base directory, but it separates them from the other Python stuff.

Also, a readme for how to build would be useful.

etpeterson (changes requested):
Looks good. I haven't tried it yet however, but it seems like it should build. I wouldn't be too worried about the build size. You could explore the alpine based ones, but that's probably not worth it.

requirements.txt
Unique-Usman: Yeah, logging is now part of python standard library.

.dockerignore
Unique-Usman: Done

Unique-Usman:
I tried out alpine image, but, it does not work as it does not have support for installing pytorch.

etpeterson:
Looks good. I actually haven't been able to test it but it's a new feature and low impact on existing things, so I think it's safe to merge. I'll try to check in the next few days though.

Unique-Usman:
Thanks @etpeterson, let's move to the goal.

PR #67, opened 2024-06-02
Add algorithm comparison and visualization dashboard

AhmedBasem20:
This PR adds an initial visualization dashboard website. I'll set this as a draft for now because I made the deployment for this to replace the documentation workflow. I'm working on creating a sub-page URL for the documentation website to be `/documentation` and the dashboard page to be on the root URL.


etpeterson:
Do you have your own branch or repo where we can see it in action?

Is it possible to add a "Loading" message because the initial load takes some time. Or maybe even a progress bar or text if you want to get fancy.

I still see that deselecting columns doesn't remove the true value point.
image

Are other plot types available, like violin?

etpeterson:
Looking at the range slider, it goes from 2x to 100x, but most of that range isn't used. Maybe it can max out at the maximum useful value?

It might be harder, but a double slider to get a value for the top and bottom of the range would be nice for a few cases - a thought, not a requirement.

It's also a little strange that there are two places where zooming can happen, the "Range" and on the plot "Zoom". It's not immediately clear how they interact and when to use which one. Is it possible to bring those together in some way?

AhmedBasem20:
> Do you have your own branch or repo where we can see it in action?

Dashboard: https://ahmedbasem20.github.io/TF2.4_IVIM-MRI_CodeCollection
Documentation: https://ahmedbasem20.github.io/TF2.4_IVIM-MRI_CodeCollection/documentation

I added a loading state, and will look into a way to remove the ground truth points when deselected.

> Are other plot types available, like violin?

Yes, the violin chart is available, there is a diverse number of chart types available here: https://plotly.com/javascript/

.github/workflows/website.yml
```diff
@@ -33,25 +33,39 @@ jobs:
...
+      - name: 'Download analysis data'
+        if: ${{ github.event.workflow_run.conclusion == 'success' }}
+        uses: ./.github/actions/download-artifact
+        with:
+          name: 'Data'
 
-      - name: Build html
+      - name: move data to the website folder
+        run: |
+          mv test_output.csv website/
```
etpeterson, line 51: I think we should probably avoid mixing build files with repo files. It looks like previously we were moving everything into a few folder and pushing from there. That's probably a slightly safer approach. Right now, for example, there's `combined_report.html` in the website folder. How does that interact with these new files?

.github/workflows/website.yml
```diff
@@ -33,25 +33,39 @@ jobs:
...
         run: |
           mkdir docs/_static
           mv *.pdf docs/_static/
           sphinx-apidoc -o docs src
           cd docs/
           make html
 
+      - name: move data to the website folder
+        run: |
+          mv docs/_build/html website/documentation
```
etpeterson, line 63: Is this a space in the folder name? Probably should avoid that for safety.

.github/workflows/website.yml
```diff
@@ -33,25 +33,39 @@ jobs:
...
+        uses: ./.github/actions/download-artifact
+        with:
+          name: 'Data'
 
-      - name: Build html
+      - name: move data to the website folder
+        run: |
+          mv test_output.csv website/
+
+      - name: Build documentation
```
etpeterson, line 53: Maybe I'm not understanding the new folder structure here but it seems like the documentation is getting lost because there's not link on the home page. Maybe there should be a new home that links to everything (dashboard, old figures, documentation)?

etpeterson (changes requested):
Did you also have changes coming for removing the points when hiding the group? Also the range changes we talked about? Or is that for a later PR?

.github/workflows/website.yml
AhmedBasem20: What do you think of this one? a bit simple and empty but we can fill it with some content from the README
image

.github/workflows/website.yml
AhmedBasem20: done

.github/workflows/website.yml
AhmedBasem20: I created a separate folder `dashboard_website` where only the code needed for building the dashboard resides.

etpeterson:
I noticed some force pushes here. Why is that necessary? Best to avoid if possible.

.github/workflows/website.yml
etpeterson: That would work. Maybe less text - it says OSIPI Taskforce 2.4 3 times in that page. That could probably just be in the title. Is this active on your site? It looks like it's still the previous version.

dashboard_website/index.js
```diff
@@ -0,0 +1,419 @@
...
+        let newValue = parseInt(rangeSlider.value) - 2;
+        if (newValue >= 2) {
+            rangeSliderLower.value = newValue;
+            updateRangeLower(newValue);
+        }
+    });
+
+    incrementRangeLower.addEventListener('click', function() {
+        let newValue = parseInt(rangeSliderLower.value) + 2;
+        if (newValue <= rangeSliderLower.max) {
```
etpeterson, line 116: Sometimes this prevents the increment from getting to the maximum value. If the click pushes it above the max it does nothing. It should just set it to the max.

dashboard_website/index.js
```diff
@@ -0,0 +1,419 @@
...
+    decrementRangeLower.addEventListener('click', function() {
+        let newValue = parseInt(rangeSlider.value) - 2;
+        if (newValue >= 2) {
+            rangeSliderLower.value = newValue;
+            updateRangeLower(newValue);
+        }
+    });
+
+    incrementRangeLower.addEventListener('click', function() {
+        let newValue = parseInt(rangeSliderLower.value) + 2;
```
etpeterson, line 115: Why 2 here?

dashboard_website/index.js
```diff
@@ -0,0 +1,419 @@
...
+        let newValue = parseInt(rangeSliderRegion.value) + 2;
+        if (newValue <= rangeSliderRegion.max) {
+            rangeSliderRegion.value = newValue;
+            updateRangeRegion(newValue);
+        }
+    });
+
+    // Add event listeners to range slider and buttons for region
+    const rangeSliderLowerRegion = document.getElementById('lower-range-slider-region');
+    const rangeValueLowerRegion = document.getElementById('lower-range-value-region');
```
etpeterson, line 184: This seems to be inverted from what I would expect. Clicking `+` on the lower slider decreases the minimum rather than increasing it.

Also, I noticed a case where the max and min were the same value so clicking it didn't do anything. Maybe it could be disabled in those cases?

etpeterson:
I took the number to be the current value. But it actually seems to be the maximum value. If that's the case there should also be the similar number on the other side.
image

etpeterson (changes requested):
Looking good but I found some quirks.

dashboard_website/index.js
AhmedBasem20: > This seems to be inverted from what I would expect. Clicking + on the lower slider decreases the minimum rather than increasing it.

Fixed. I thought it made sense this way because clicking `+` expands the view for more data points.

> Also, I noticed a case where the max and min were the same value so clicking it didn't do anything. Maybe it could be disabled in those cases?

Done.

dashboard_website/index.js
AhmedBasem20: It increments the range slider when clicking on `+` button by 2 steps, because 1 is a bit slow.

.github/workflows/website.yml
AhmedBasem20: Done. It was not merged yet there but now it is active: https://ahmedbasem20.github.io/TF2.4_IVIM-MRI_CodeCollection/

AhmedBasem20:
Thanks for the review @etpeterson
I made some improvements based on the points you mentioned, please take a look again and let me know what you think.

etpeterson:
> I took the number to be the current value. But it actually seems to be the maximum value. If that's the case there should also be the similar number on the other side. image

Thoughs on this?

etpeterson:
@oliverchampion This is looking good to me. Did you have comments or thoughts before merge?

oliverchampion:
Looks good! Just some remarks:
At some settings the boundaries seem to get stuck. E.g.:
Knipsel
I cannot do anything with the lower boundary. 
If easy to fix, than yes please, but if challenging, I think the GUI works good enough and one can always zoom using the built in functions :).

AhmedBasem20:
Hey Oliver, this bug appears when the range is very small and the full range is already visible. I attempted to debug it but it seems a bit challenging to resolve.

etpeterson (approved):
Looks good to me.

etpeterson:
@oliverchampion This looks good to me. Feel free to merge if you're satisfied.

PR #71, opened 2024-07-24
Abdominal phantom

oliverchampion:

The abdominal phantom has been updated; we need to check whether this influences the automated tests from Eric

### Checklist

- [ x ] Self-review of changed code
- [ x ] Added automated tests where applicable
- [ x ] Update Docs & Guides


phantoms/MR_XCAT_qMRI/sim_ivim_sig.py
```diff
@@ -261,7 +273,7 @@ def XCAT_to_MR_DCE(XCAT, TR, TE, bvalue, D, f, Ds, b0=3, ivim_cont = True):
         73: 'Pancreas tumor',
     }
     ###############################################################################
-    np.random.seed(42)
+    np.random.seed(41)
```
etpeterson, line 276: Seed change?

utilities/data_simulation/Download_data.py
```diff
@@ -2,7 +2,7 @@
 import os
 import subprocess
 import time
-
+import zenodo_get
```
etpeterson, line 5: Is this used?

etpeterson:
I think the remaining piece is to run sim_ivim_sig.py to get the new generic.json and update that. Hopefully the tests will work!

phantoms/MR_XCAT_qMRI/sim_ivim_sig.py
oliverchampion: This is now fixed. Changed it to test something :).

utilities/data_simulation/Download_data.py
oliverchampion: It is used here:    

subprocess.check_call(["zenodo_get", 'https://zenodo.org/records/10696605'])

but as supprocess, it is not recognized that it is used. If you remove it, however, this crashes :).

etpeterson: I don't understand that, but I guess it's not a big deal.

etpeterson:
> I think the remaining piece is to run sim_ivim_sig.py to get the new generic.json and update that. Hopefully the tests will work!

I think this is the remaining piece?

utilities/data_simulation/Download_data.py
oliverchampion: I think this starts a subprocess which then runs the commando in the string (which contains Zenodo_get). But the Zenodo_get function is not recognized by interpreters as it is hidden in a string.
The download happens outside Pycharm, and if I don't put this in a subprocess, it somehow continues the code while downloading, running into errors (using data that still needs downloading).

etpeterson:
> > I think the remaining piece is to run sim_ivim_sig.py to get the new generic.json and update that. Hopefully the tests will work!
> 
> Ah, I had run this, but the generic.json files are on the .gitignore list so nothing was pushed. Is there a good way of pushing these results? Also, it creates only four "generic_#.json"-files (with # one-four), but not a "generic.json" file. Are those files manually copied to tests/IVI-models/unit_tests? Which files are used and which of these is (renamed into?) "generic.json"?
> 
> Thanks!

Ah, there was a change. I think there should also be a `generic_original.json`? I think it's being pulled from `phantoms/MR_XCAT_qMRI/b_values.json` now. I think you should already have that, so rename that generic.json and overwrite the original and `git add -f generic.json`. The `-f` forces it to add even though it's ignored.

Then hopefully it all works!

etpeterson (changes requested):
Just the generic.json overwrite remaining.

oliverchampion:
Good news: File is updated (as well as the other generic files). 
Bad news: (some) tests fail... But a lot pass... Will need to check this in more detail tomorrow...

oliverchampion:
@etpeterson It seems like only your algorithm is failing, and only in a limited set of tissues. In particular, it has trouble estimating an f of 1 in vessels. But also has some challenges in the pericardium...

etpeterson:
> @etpeterson It seems like only your algorithm is failing, and only in a limited set of tissues. In particular, it has trouble estimating an f of 1 in vessels. But also has some challenges in the pericardium...
> 
> I think it has to do with the fact that all algorithms use real tolerances, but yours uses dynamic tolerances. Hence, your tolerances are way more strict (and realistic) than for the other algorithms.
> 
> I think the dyn_tolerances are related to an issue/feature request I wrote earlier today :) and should be the way to go.

I'm not too sure why it's failing now, but I'm not too worried. You could mark those tests as expected failures. So add the regions to this area of algorithms.json. That's probably easiest, especially since it seems like more bounds changes are coming.

```
"ETP_SRI_LinearFitting": {
        "xfail_names": {
            "Vein": true
        },
```

oliverchampion:
Features updated:
1: Phantom is smaller (done previously)
2: signal in phantom can now optionally be T1/T2 weighted. Default has T1/T2 weighting turned off. 

> > @etpeterson It seems like only your algorithm is failing, and only in a limited set of tissues. In particular, it has trouble estimating an f of 1 in vessels. But also has some challenges in the pericardium...
> > I think it has to do with the fact that all algorithms use real tolerances, but yours uses dynamic tolerances. Hence, your tolerances are way more strict (and realistic) than for the other algorithms.
> > I think the dyn_tolerances are related to an issue/feature request I wrote earlier today :) and should be the way to go.
> 
> I'm not too sure why it's failing now, but I'm not too worried. You could mark those tests as expected failures. So add the regions to this area of algorithms.json. That's probably easiest, especially since it seems like more bounds changes are coming. Actually, looks like Vein was failing and it's now passing, so just swap out "Blood RV" for "Vein" and it should work.
> 
> ```
> "ETP_SRI_LinearFitting": {
>         "xfail_names": {
>             "Vein": true
>         },
> ```

Ah! Now I understand the XPASS fail, haha. I was wondering what that was doing :)

tests/IVIMmodels/unit_tests/test_ivim_fit.py
```diff
@@ -169,6 +169,8 @@ def test_ivim_fit_saved(name, bvals, data, algorithm, xfail, kwargs, tolerances,
     tolerances = tolerances_helper(tolerances, ratio, data["noise"])
     [f_fit, Dp_fit, D_fit] = fit.osipi_fit(signal, bvals)
     npt.assert_allclose(data['f'], f_fit, rtol=tolerances["rtol"]["f"], atol=tolerances["atol"]["f"])
-    npt.assert_allclose(data['D'], D_fit, rtol=tolerances["rtol"]["D"], atol=tolerances["atol"]["D"])
-    npt.assert_allclose(data['Dp'], Dp_fit, rtol=tolerances["rtol"]["Dp"], atol=tolerances["atol"]["Dp"])
+    if data['f']<0.80: # we need some signal for D to be detected
```
etpeterson, line 172: Just looking at code in more detail, would this be a relative tolerance thing rather than an `if`?

etpeterson (changes requested):

tests/IVIMmodels/unit_tests/test_ivim_fit.py
oliverchampion: Yes, I am doing that in my next project. But either way it seems silly to quantity something that is undefined (D* is f=0 and D if f=1)

etpeterson: Understood. Looks good to me.

etpeterson (approved):

PR #72, opened 2024-07-24
Optimize bandwidth

AhmedBasem20:
Optimize data handling: Implement rounding and gzip compression for CCV output

- Implement rounding to four decimal places for specific numerical columns.
- gzip compression for CSV outputs to reduce file size and improve bandwidth usage during data transfers.

The output file size is now ~20MB. If we dropped the bvalues columns (which are not used for visualization currently) the file will be only just 2.5MB!


oliverchampion:
Great idea Ahmed. But, also @etpeterson:
Maybe instead of solving this ad hoc, we should reconsider what tests need to be run on the fly each time (and saved).
- Does it not make more sense to reduce the digit size when generating the csv files already?
- Do we need all the b-value testing to be done for all voxels at all SNRs? Should all of this be run and 20MB be generated every time we make a pull request? Not sure whether this should be separate from unit testing?
- Can we save the signal per b-value separately? And possibly even the reference values. It seems now to save a lot of redundant data.

Possible alternative:
- a Json file makes life easier? 
- 2 excel files (input and output)
- Alternatively, Ahmed pre-processes the excel and saves a copy that is easier to read.

oliverchampion:
Moreover, we need the unit tests to succeed :)

tests/IVIMmodels/unit_tests/reduce_output_size.py
```diff
@@ -0,0 +1,18 @@
+import pandas as pd
+
+df = pd.read_csv('test_output.csv')
```
etpeterson, line 3: This is particularly inefficient because it loads everything and operates on it all in memory at once. This is a big file so it might be worth either fixing this now or opening an issue to do this later.

An optimization would be to not use pandas and use the simpler csv and operate line by line. Python can still handle the zipping or it could be the linux utility.

tests/IVIMmodels/unit_tests/reduce_output_size.py
etpeterson, line 1: Seems like we aren't dropping any columns?

etpeterson (changes requested):
Going in the right direction but fix the tests!

tests/IVIMmodels/unit_tests/reduce_output_size.py
AhmedBasem20: Done.

tests/IVIMmodels/unit_tests/reduce_output_size.py
AhmedBasem20: Fixed in the new version (using the native csv module)

utilities/reduce_output_size.py
```diff
@@ -0,0 +1,26 @@
...
+        # Drop b_values columns
+        fieldnames = [field for field in reader.fieldnames if not field.startswith('bval_')]
+        writer = csv.DictWriter(outfile, fieldnames=fieldnames)
+        writer.writeheader()
+
+        columns_to_round = ['f', 'Dp', 'D', 'f_fitted', 'Dp_fitted', 'D_fitted']
+        
+        # Process each row
+        for row in reader:
+            filtered_row = {column: row[column] for column in fieldnames}
```
etpeterson, line 22: Minor optimization, but it's probably faster to delete the columns you don't want rather than rebuilding the dictionary with columns you do want.

etpeterson (changes requested):

utilities/reduce_output_size.py
AhmedBasem20: I've made the requested changes. Could you please take another look and let me know if it is good to go?

etpeterson (approved):
Looks good to me

PR #73, opened 2024-08-07
Save test results to a json file & Display results Visuals on the website

AhmedBasem20:


AhmedBasem20:
Hello @etpeterson, @oliverchampion
I managed to capture the expected and actual fitting values for each test and store them as a json file.
You can check [the website here](https://ahmedbasem20.github.io/TF2.4_IVIM-MRI_CodeCollection/dashboard/) to see the results, please let me know if you have any comments, thanks.

Data point example:
```
         {
            "name": "Myocardium LV",
            "algorithm": "ETP_SRI_LinearFitting",
            "f_fit": 0.1851456126831551,
            "Dp_fit": 0.007000571619495537,
            "D_fit": 0.0023544011841776917,
            "f": 0.14999999999999997,
            "Dp": 0.08,
            "D": 0.0024,
            "status": "PASSED"
        }
```

Tolerance values:

```
    "rtol": {
        "f": 5,
        "D": 5,
        "Dp": 25
    },
    "atol": {
        "f": 0.01,
        "D": 0.01,
        "Dp": 0.1
    }
```

tests/IVIMmodels/unit_tests/test_ivim_fit.py
```diff
@@ -161,14 +62,35 @@ def data_ivim_fit_saved():
...
+        "f_fit": to_list_if_needed(f_fit),
+        "Dp_fit": to_list_if_needed(Dp_fit),
+        "D_fit": to_list_if_needed(D_fit),
+        "f": to_list_if_needed(data['f']),
+        "Dp": to_list_if_needed(data['Dp']),
+        "D": to_list_if_needed(data['D']),
+        "status": "PASSED"
+    }
+    if xfail["xfail"]:
+        test_result['status'] = "XFAILED"
```
etpeterson, line 88: What happens if this test fails? I'm not sure actually if it would continue to run or stop, but the `test_result` clearly only has `PASS` and `XFAILED`.

tests/IVIMmodels/unit_tests/test_ivim_fit.py
```diff
@@ -161,14 +62,35 @@ def data_ivim_fit_saved():
...
+
+    test_results.append(test_result)
+    with open('test_results_report.json', 'w') as f:
```
etpeterson, line 91: This is writing the entire json for each test which is surely slowing things down a lot. It's also custom code that's going to need to be considered for every change of the test.

What I suggested last time and I'll suggest again is to look into other ways pytest can do this. A 5 minute google search is showing `pytest --json-report`, `Pytest-harvest`, and possibly pytest fixtures. Have you considered any of these?

etpeterson (changes requested):
I think the logging needs to be changed in some way.

tests/IVIMmodels/unit_tests/test_ivim_fit.py
AhmedBasem20: I did look into pytest-json-report but It gives a complex report with unnecessary information, the only way I see to include our custom values is by logging it into the code, and the final result file is not clean after then.

I'm now using pytest fixture for handling report generation.

tests/IVIMmodels/unit_tests/test_ivim_fit.py
AhmedBasem20: For the current setup it fails on the CI, is that correct? Or should it be marked on the file as `Failed` and the test should pass?
I tried changing the tolerance values and it fails for some cases with that error for example:
```
E           AssertionError: 
E           Not equal to tolerance rtol=0.05, atol=0.01
E           
E           Mismatched elements: 1 / 1 (100%)
E           Max absolute difference: 0.03707215
E           Max relative difference: 0.3462352
E            x: array(0.07)
E            y: array(0.107072)
```
image

etpeterson: I don't think the test should be failing currently. If you changed it to fail then that makes sense.

But if it fails does it write the file? And if it does write the file it doesn't correctly report the failed test because the only two options for `test_result['status']` are `PASSED` and `XFAILED`. This is why I'm strongly suggesting using a package or some method that is designed to do this rather than trying to force something.

I'm also noticing that you're not doing anything with this value. Ideally the plots would have colored points based on the status, for example green for pass, blue for xfail, red for fail.

tests/IVIMmodels/unit_tests/test_ivim_fit.py
etpeterson: What about `record_property`? https://docs.pytest.org/en/7.1.x/how-to/output.html

Not sure what you mean about the final result file not being clean, but using what is essentially a global variable isn't great. You essentially created a custom fixture that no other test can use because doing so would corrupt the test logging. A "manual" logging solution like this should be the last resort - it's very hard to maintain and because it operates outside of the testing structure it's hard to keep consistent, e.g. capturing the pass/xfail/fail criteria.

etpeterson (changes requested):

tests/IVIMmodels/unit_tests/test_ivim_fit.py
AhmedBasem20: I got your point, thanks for explaining.
image
That's how the report looks like, I used `pytest --json-report` with `record_property` for capturing the data. I'll update the frontend depending on that report and update the PR.

etpeterson: Ok, just from that it doesn't look too bad. Seems like the code changes are still incoming?

AhmedBasem20: yeah I'm getting weird bugs on the frontend after updating code, I'll give it a final attempt and then push.
image

.github/workflows/website.yml
```diff
@@ -46,12 +46,19 @@ jobs:
         with:
           name: 'Data'
 
+      - name: Run the test that generates the plots report.
+        run: |
+          pytest tests/IVIMmodels/unit_tests/test_ivim_fit.py --json-report
```
etpeterson, line 51: I think `json-report` is now a requirement?

Also, why run this test again and not get the same data from the initial run?

tests/IVIMmodels/unit_tests/test_ivim_fit.py
```diff
@@ -4,113 +4,23 @@
...
 import os
-
+import logging
 from src.wrappers.OsipiBase import OsipiBase
 from utilities.data_simulation.GenerateData import GenerateData
-
+logger = logging.getLogger(__name__)
 #run using python -m pytest from the root folder
 
+@pytest.fixture(scope="session")
```
etpeterson, line 13: Is this all still necessary?

etpeterson (changes requested):

.github/workflows/website.yml
AhmedBasem20: > I think json-report is now a requirement?

yeah, done.

> Also, why run this test again and not get the same data from the initial run?

Do you mean the unit test run? I don't find a straight way of doing it, as we have a matrix of runs so which file we should get the data from those multiple jobs?
image

tests/IVIMmodels/unit_tests/test_ivim_fit.py
AhmedBasem20: no, removed now.

website/dashboard/test_plots.js
```diff
@@ -0,0 +1,54 @@
...
+                var scatter_trace = {
+                    x: reference_values,
+                    y: errors,
+                    mode: 'markers',
+                    type: 'scatter',
+                    name: parameter.toUpperCase() + ' Fit Errors'
+                };
+
+                var tolerance_trace = {
+                    x: reference_values,
```
etpeterson, line 22: This should be data min to data max or just fixed numbers

website/dashboard/test_plots.js
```diff
@@ -0,0 +1,54 @@
...
+    fetch('report-summary.json')
+        .then(response => response.json())
+        .then(data => {
+            function createPlot(container, parameter) {
+                var reference_values = data.map(d => d[parameter]);
+                var fit_values = data.map(d => d[parameter + '_fit']);
+                var errors = fit_values.map((d, i) => d - reference_values[i]);
+
+                var tolerance = reference_values.map((d, i) => data[i]['atol'][parameter] + data[i]['rtol'][parameter] * d);
+                var negative_tolerance = reference_values.map((d, i) => -(data[i]['atol'][parameter] + data[i]['rtol'][parameter] * d));
```
etpeterson, line 11: Could simplify to negative tolerance rather than running the whole equation again.

etpeterson (changes requested):

tests/IVIMmodels/unit_tests/test_ivim_fit.py
```diff
@@ -4,114 +4,12 @@
 import json
 import pathlib
 import os
-
+import logging
 from src.wrappers.OsipiBase import OsipiBase
 from utilities.data_simulation.GenerateData import GenerateData
-
+logger = logging.getLogger(__name__)
```
etpeterson, line 10: I think logging is also still not necessary?

website/dashboard/test_plots.js
AhmedBasem20: That's how it looks like after the changes:
image

.github/workflows/website.yml
etpeterson: What about saving from the first run only? https://stackoverflow.com/questions/70532207/github-action-yaml-how-to-refer-an-item-in-a-matrix-array

But also only if it's one that runs the analysis. Maybe that's too much optimization.

website/dashboard/test_plots.js
```diff
@@ -0,0 +1,58 @@
+document.addEventListener('DOMContentLoaded', function() {
+    fetch('report-summary.json')
+        .then(response => response.json())
+        .then(data => {
+            function createPlot(container, parameter) {
+                var reference_values = data.map(d => d[parameter]);
+                var fit_values = data.map(d => d[parameter + '_fit']);
+                var errors = fit_values.map((d, i) => d - reference_values[i]);
+
+                var tolerance = reference_values.map((d, i) => data[i]['atol'][parameter] + data[i]['rtol'][parameter] * d);
```
etpeterson, line 10: This doesn't seem right to me, both the code and the plots. For example, your last plot shows the error bounds decreasing as x increases, that could only happen if `rtol<0`.

In the code you calculate the tolerance values for every single point then calculate the min and max x-value, then plot the first 2 tolerances with those min and max values? Shouldn't it be the other way around? Calculate the min and max x-values and then calculate the y tolerances?

etpeterson (changes requested):

website/dashboard/test_plots.js
AhmedBasem20: Just pushed the changes, here is the last update:
image

website/dashboard/test_plots.js
```diff
@@ -6,14 +6,14 @@ document.addEventListener('DOMContentLoaded', function() {
...
-
-                var tolerance = reference_values.map((d, i) => data[i]['atol'][parameter] + data[i]['rtol'][parameter] * d);
-                var negative_tolerance = tolerance.map(t => -t);
-
                 var minRefValue = Math.min(...reference_values);
                 var maxRefValue = Math.max(...reference_values);
                 // Create a range for tolerance trace x axis.
                 var xRange = [minRefValue, maxRefValue];
+
+                var tolerance = xRange.map((d, i) => data[reference_values.indexOf(d)]['atol'][parameter] + data[i]['rtol'][parameter] * d);
```
etpeterson, line 14: Your index in `rtol` should be the same as in `atol`. You're choosing different datasets for both parts of the equation.

etpeterson, line 14: I think we had also talked about plotting one or multiple bounds? What do you think @oliverchampion Currently this is somewhat randomly choosing the bounds as the bounds of the values at the endpoints.

etpeterson (changes requested):

website/dashboard/test_plots.js
oliverchampion: @AhmedBasem20, do you have the link to the latest dashboard?
@etpeterson For unit testing (errors), I think having 1 bound is probably okay for now. When we go into functional testing, we could add f-dependent and SNR-dependent bounds, but I'm also unsure how one would intuitively visualize multiple bounds...

etpeterson: So the question is, what bound should we choose? When fixed it will be whatever datapoint is at the beginning or end. That's somewhat random so should it be the max or min y-extent?

AhmedBasem20: yeah, that's how the plots look after correctly accessing the tolerances of the min and max reference values:

image

AhmedBasem20: > @AhmedBasem20, do you have the link to the latest dashboard?

I'm deploying it, it will be available in ~30 minutes here: https://ahmedbasem20.github.io/TF2.4_IVIM-MRI_CodeCollection/dashboard/

etpeterson: That error was something that I've been meaning to fix. I think it should be fixed here, but it will take a while to see. So you can merge that into yours if it's working. https://github.com/OSIPI/TF2.4_IVIM-MRI_CodeCollection/pull/82/

AhmedBasem20: it is now working [here](https://ahmedbasem20.github.io/TF2.4_IVIM-MRI_CodeCollection/dashboard/), please let me know what you think, thanks.

AhmedBasem20:
Hey @etpeterson @oliverchampion Please check the following [dashboard updates](https://ahmedbasem20.github.io/TF2.4_IVIM-MRI_CodeCollection/dashboard/):
- I added default tolerance bounds that accurately wrap the data after visual testing.
- Set data point color depending on the test status. (passed, failed, or xfailed).
- Display algorithm and region when hovering over data points.

etpeterson (approved):

PR #74, opened 2024-08-11
Added a yml file to build the docker image

Unique-Usman:
### Create a github action for building and running the docker image. 

@etpeterson, hello, how are you doing and how have you been ?

Should we make the github action to build the image and also run it or just building. For running, we will need some data, should we download the data during in the github action or add the data as part of the repository ?


.github/workflows/docker-build-and-run.yml
```diff
@@ -0,0 +1,33 @@
...
+
+      # Build the Docker image
+      - name: Build Docker image
+        run: |
+          docker build -t tf2.4_ivim-mri_codecollection -f Docker/Dockerfile .
+
+      # Run the Docker container
+      - name: Run Docker container
+        run: |
+          docker run -it --rm --name TF2.4_IVIM-MRI_CodeCollection \
```
etpeterson, line 30: The check failed, it's saying the terminal is not a TTY, so I think drop the `-it` or just the `t` part of that?

.github/workflows/docker-build-and-run.yml
```diff
@@ -0,0 +1,33 @@
+name: Build and Run Docker
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
```
etpeterson, line 7: It's set to run for all pull requests. The docker build was very fast, but the run part won't be. But I'm also thinking not every pull request to main needs to build the docker image. Maybe just the main branch itself? Or, like the analysis pipeline, the branches that start with `docker/**` could also build it?

.github/workflows/docker-build-and-run.yml
```diff
@@ -0,0 +1,33 @@
...
+        run: |
+          docker build -t tf2.4_ivim-mri_codecollection -f Docker/Dockerfile .
+
+      # Run the Docker container
+      - name: Run Docker container
+        run: |
+          docker run -it --rm --name TF2.4_IVIM-MRI_CodeCollection \
+            -v ${{ github.workspace }}:/usr/src/app \
+            -v ${{ github.workspace }}:/usr/app/output \
+            tf2.4_ivim-mri_codecollection brain.nii.gz brain.bvec brain.bval
```
etpeterson, line 33: I don't think the data are available. We could pull it, but that run would take a while. Maybe we could read a small text file that does just a few fits and takes just a second or two. We really just want to run it to confirm it works, we have other tests to check the actual results.

.github/workflows/docker-build-and-run.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      # Build the Docker image
+      - name: Build Docker image
+        run: |
+          docker build -t tf2.4_ivim-mri_codecollection -f Docker/Dockerfile .
```
etpeterson, line 25: Is this docker image getting saved as an artifact? I'm not sure how that works in github actions but it's probably possible. At the very least I know it's possible to dump docker images to files.

etpeterson (changes requested):
I really like this direction. Just a few comments but I think it's really close already.

etpeterson:
> ### Create a github action for building and running the docker image.
> @etpeterson, hello, how are you doing and how have you been ?
> 
> Should we make the github action to build the image and also run it or just building. For running, we will need some data, should we download the data during in the github action or add the data as part of the repository ?

Doing well and thanks for coming back to this. I submitted a review for you.

.github/workflows/docker-build-and-run.yml
Unique-Usman: I save the image as an artifact now.

.github/workflows/docker-build-and-run.yml
Unique-Usman: How to go about the reading the small text file ?

.github/workflows/docker-build-and-run.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build Docker image
+        run: |
+          docker build -t tf2.4_ivim-mri_codecollection -f Docker/Dockerfile .
+
+      - name: Save Docker image to a tarball
+        run: |
+          docker save -o tf2.4_ivim-mri_codecollection.tar tf2.4_ivim-mri_codecollection
```
etpeterson, line 30: These things are large, how about also zipping it.

etpeterson (changes requested):

Unique-Usman:
@etpeterson

https://github.com/OSIPI/TF2.4_IVIM-MRI_CodeCollection/blob/main/utilities/data_simulation/GenerateData.py#L25

https://github.com/OSIPI/TF2.4_IVIM-MRI_CodeCollection/blob/main/tests/IVIMmodels/unit_tests/test_ivim_synthetic.py#L23C4-L23C34

How do we get the ivim_data ?

etpeterson:
> @etpeterson
> 
> https://github.com/OSIPI/TF2.4_IVIM-MRI_CodeCollection/blob/main/utilities/data_simulation/GenerateData.py#L25
> 
> https://github.com/OSIPI/TF2.4_IVIM-MRI_CodeCollection/blob/main/tests/IVIMmodels/unit_tests/test_ivim_synthetic.py#L23C4-L23C34
> 
> How do we get the ivim_data ?

I'm just saying random but realistic f [0, 1], D [0, 1e-3], D* [0, 1e-1], generate b-value curves from that using the GenerateData that you linked to and the b-values I suggested above. Generate that volume and use that as a test dataset.

Unique-Usman:
@etpeterson, I think I might need your regarding this part. Thank you. I am able to get it.

etpeterson:
> @etpeterson, I think I might need your regarding this part. Thank you. I am able to get it.

Something roughly like this. I haven't tested it but it has the basics.

```
gd = GenerateData()
f_in = np.random.uniform(low=0, high=1, size=(10, 10, 5))
D_in = np.random.uniform(low=0, high=1e-3, size=(10, 10, 5))
Dp_in = np.random.uniform(low=0, high=1e-1, size=(10, 10, 5))
bvals = [0, 50, 100, 500, 1000]
images = gd.ivim_signal(self, D_in, Dp_in, f_in, S0, bvalues)
save_nii(images)
```

Unique-Usman:
> > @etpeterson, I think I might need your regarding this part. Thank you. I am able to get it.
> 
> Something roughly like this. I haven't tested it but it has the basics.
> 
> ```
> gd = GenerateData()
> f_in = np.random.uniform(low=0, high=1, size=(10, 10, 5))
> D_in = np.random.uniform(low=0, high=1e-3, size=(10, 10, 5))
> Dp_in = np.random.uniform(low=0, high=1e-1, size=(10, 10, 5))
> bvals = [0, 50, 100, 500, 1000]
> images = gd.ivim_signal(self, D_in, Dp_in, f_in, S0, bvalues)
> save_nii(images)
> ```

Did something similay earlier. Apparently the error seems to come from GenerateData due to shape imcompability.

https://github.com/OSIPI/TF2.4_IVIM-MRI_CodeCollection/blob/main/utilities/data_simulation/GenerateData.py#L56

https://github.com/OSIPI/TF2.4_IVIM-MRI_CodeCollection/blob/main/utilities/data_simulation/GenerateData.py#L141

etpeterson:
> > > @etpeterson, I think I might need your regarding this part. Thank you. I am able to get it.
> > 
> > 
> > Something roughly like this. I haven't tested it but it has the basics.
> > ```
> > gd = GenerateData()
> > f_in = np.random.uniform(low=0, high=1, size=(10, 10, 5))
> > D_in = np.random.uniform(low=0, high=1e-3, size=(10, 10, 5))
> > Dp_in = np.random.uniform(low=0, high=1e-1, size=(10, 10, 5))
> > bvals = [0, 50, 100, 500, 1000]
> > images = gd.ivim_signal(self, D_in, Dp_in, f_in, S0, bvalues)
> > save_nii(images)
> > ```
> 
> Did something similay earlier. Apparently the error seems to come from GenerateData due to shape imcompability.
> 
> https://github.com/OSIPI/TF2.4_IVIM-MRI_CodeCollection/blob/main/utilities/data_simulation/GenerateData.py#L56
> 
> https://github.com/OSIPI/TF2.4_IVIM-MRI_CodeCollection/blob/main/utilities/data_simulation/GenerateData.py#L141

What's the error specifically? I'm not sure if I've ever tried GenerateData with multiple values at a time, but it doesn't seem like the lines you linked should be causing problems.

Unique-Usman:
Screenshot from 2024-08-30 06-12-39

Unique-Usman:
assert D >= 0, 'D must be >= 0'

etpeterson:
>  assert D >= 0, 'D must be >= 0' 

Ok, so this assert is correct except that it should be in bumpy format, so `assert np.all(D>=0)`.

Unique-Usman:
So, I made this changes, but, it seems there are multiple bugs in the GenerateData.py.

Screenshot from 2024-09-01 23-13-25

etpeterson:
> So, I made this changes, but, it seems there are multiple bugs in the GenerateData.py.
> Screenshot from 2024-09-01 23-14-49
> 
> 

Could be, but hard to tell just from that. Your inputs are all the same size?

etpeterson:
Hey, just checking - are you still working on this?

Unique-Usman:
@etpeterson, I have resumed working on it, should get back in few minutes.

Unique-Usman:
@etpeterson kindly go through this whenever you are chanced.

utilities/data_simulation/np
etpeterson, line 1: What is this file?

Docker/generate_signal_docker_test.py
```diff
@@ -0,0 +1,31 @@
+import numpy as np
+import nibabel as nib
+from utilities.data_simulation.GenerateData import GenerateData
+
+def save_nii(data, filename='ivim_image.nii.gz'):
```
etpeterson, line 5: There's already a `save_nifti_file` in the wrapper.

Docker/generate_signal_docker_test.py
etpeterson, line 1: What is the purpose of this file? It looks like it generates an image but I don't see it being used for anything. I was expecting to see the simulated image and associated files generated by this.

Docker/ivim_simulation.nii.gz
etpeterson, line 1: This and the associated bval and bvec files are inside the docker folder, so they're getting added to the docker image. We don't want that. Also, shouldn't they be generated on the fly rather than be pre-generated?

.github/workflows/docker-build-and-run.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: Upload Docker image artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: docker-image
+          path: tf2.4_ivim-mri_codecollection.tar.gz
+
+      - name: Run Docker container
+        run: |
+          docker run --rm --name TF2.4_IVIM-MRI_CodeCollection \
```
etpeterson, line 40: We should confirm that images are actually generated. Maybe not the exact values in the images, but even that the files we expect exist.

etpeterson (changes requested):
Looking good. A few more comments.

utilities/data_simulation/np
Unique-Usman: Some erroneus files, will remove it.

Docker/generate_signal_docker_test.py
Unique-Usman: Yeah, sure.

Docker/generate_signal_docker_test.py
Unique-Usman: This is the code I used for generating the signal which I was using for the docker.

Docker/ivim_simulation.nii.gz
Unique-Usman: I will generate it on the fly, I thought it makes sense having it constant.

.github/workflows/docker-build-and-run.yml
Unique-Usman: Yeah, exactly.

Unique-Usman:
@etpeterson, kindly take a look.

.github/workflows/docker-build-and-run.yml
```diff
@@ -0,0 +1,85 @@
...
+              echo "Error: $file not found"
+              exit 1
+            fi
+          done
+          echo "All input files generated successfully"
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build Docker image
```
etpeterson, line 45: I think docker buildx can build directly to files `docker buildx build --output type=tar,dest=my-image.tar .`

.github/workflows/docker-build-and-run.yml
```diff
@@ -0,0 +1,85 @@
...
+        run: |
+          docker build -t tf2.4_ivim-mri_codecollection -f Docker/Dockerfile .
+
+      - name: Save and compress Docker image 
+        run: |
+          docker save tf2.4_ivim-mri_codecollection | gzip > tf2.4_ivim-mri_codecollection.tar.gz
+
+      - name: Upload Docker image artifact
```
etpeterson, line 53: This should move down below the verification because we don't want bad docker images pushed.

.github/workflows/docker-build-and-run.yml
```diff
@@ -0,0 +1,85 @@
...
+        uses: actions/upload-artifact@v4
+        with:
+          name: docker-image
+          path: tf2.4_ivim-mri_codecollection.tar.gz
```
etpeterson, line 57: I'm not seeing why, but the artifact is empty. I wonder if it's too big? If there's no way to save it perhaps it could be split or maybe we just forget about saving the file for now.

etpeterson (changes requested):
Looking good, really just those few changes and the question about the docker file artifact.

.github/workflows/docker-build-and-run.yml
Unique-Usman: Thanks for this.

.github/workflows/docker-build-and-run.yml
Unique-Usman: Yeah, done.

.github/workflows/docker-build-and-run.yml
Unique-Usman: I will check this and verify.

Unique-Usman: @etpeterson, I think it is there, where are you checking 
Screenshot From 2024-10-01 05-29-09

.github/workflows/docker-build-and-run.yml
Unique-Usman: Yeah, it works, but, it directly save the docker image into the .tar file and we have some following line which uses the docker image and needs to be loaded again. I felt it is more cleaner having it like this.

.github/workflows/docker-build-and-run.yml
etpeterson: A whole docker image is not 198 bytes. If you download it and unzip it you'll find it empty. Also: 
image

Unique-Usman: @etpeterson, any suggested solution ?

etpeterson: You could try a cleaning 
` - name: Cleanup Docker
     run: |
       docker system prune -a --force`

You could also try alpine linux base. I think it's smaller but might not work correctly.

If neither of those works then let's build and test but not push.

Unique-Usman:
@etpeterson, both approach did not work. When I did the cleanup, I was still getting empty files. Also, the alpine base does not support the use of torch which is part of the requirement. So, I am removing the push component

etpeterson:
> @etpeterson, both approach did not work. When I did the cleanup, I was still getting empty files. Also, the alpine base does not support the use of torch which is part of the requirement. So, I am removing the push component

Ok, seems like that's not possible then. You're still trying to upload the docker image though. 
image


Also, please don't rewrite the history like you did with force pushes. It make it really hard to review because I can't see what you had before and what my previous comments relate to.

Unique-Usman:
> > @etpeterson, both approach did not work. When I did the cleanup, I was still getting empty files. Also, the alpine base does not support the use of torch which is part of the requirement. So, I am removing the push component
> 
> Ok, seems like that's not possible then. You're still trying to upload the docker image though. image
> 
My bad, I thought I removed this. I should have verfied in the actions. 
> Also, please don't rewrite the history like you did with force pushes. It make it really hard to review because I can't see what you had before and what my previous comments relate to.
I am sorry for this, I was avoiding this also.

etpeterson (approved):

PR #80, opened 2024-09-03
Multiprocess fitting script

etpeterson:

This adds multiprocess fitting to the nifti fitting script. Perhaps better would be to add it to the wrapper. This is mostly an example. It could be added but is mostly opened for suggestions.

### Checklist

- [ ] Update Docs & Guides

### DIFF
--- a/WrapImage/nifti_wrapper.py
+++ b/WrapImage/nifti_wrapper.py
@@ -1,64 +1,13 @@
 import argparse
-import json
-import os
-import nibabel as nib
+from utilities.process.file_io import read_nifti_file, read_bval_file, read_bvec_file, save_nifti_file
+from utilities.process.diffusion_utils import find_directions, find_shells, normalize_series
 from src.wrappers.OsipiBase import OsipiBase
 import numpy as np
 from tqdm import tqdm
+from tqdm.contrib.concurrent import process_map
+from functools import partial
 
 
-def read_nifti_file(input_file):
-    """
-    For reading the 4d nifti image
-    """
-    nifti_img = nib.load(input_file)
-    return nifti_img.get_fdata(), nifti_img.header
-
-def read_json_file(json_file):
-    """
-    For reading the json file
-    """
-
-    if not os.path.exists(json_file):
-        raise FileNotFoundError(f"File '{json_file}' not found.")
-
-    with open(json_file, "r") as f:
-        try:
-            json_data = json.load(f)
-        except json.JSONDecodeError as e:
-            raise ValueError(f"Error decoding JSON in file '{json_file}': {e}")
-
-    return json_data
-
-def read_bval_file(bval_file):
-    """
-    For reading the bval file
-    """
-    if not os.path.exists(bval_file):
-        raise FileNotFoundError(f"File '{bval_file}' not found.")
-
-    bval_data = np.genfromtxt(bval_file, dtype=float)
-    return bval_data
-
-def read_bvec_file(bvec_file):
-    """
-    For reading the bvec file
-    """
-    if not os.path.exists(bvec_file):
-        raise FileNotFoundError(f"File '{bvec_file}' not found.")
-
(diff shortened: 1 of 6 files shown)


WrapImage/nifti_wrapper.py
```diff
@@ -75,52 +24,162 @@ def loop_over_first_n_minus_1_dimensions(arr):
         flat_view = arr[idx].flatten()
         yield idx, flat_view
 
+def generate_data(data, bvals, b0_indices, groups, total_iteration):
```
oliverchampion, line 27: 1: Is b0_indices something you want to tell the algorithm, or is it more fool proof if we generate them locally from bvals?
2: Maybe we need to start documenting some code; for example, to me "generate_data" sounds like it will generate data, but it is unclear why it needs data for this (if there is already data, why generate it?)

WrapImage/nifti_wrapper.py
```diff
@@ -75,52 +24,162 @@ def loop_over_first_n_minus_1_dimensions(arr):
...
+    num_directions = groups.shape[1]
+    data = data.reshape(total_iteration, -1)
+    for idx in range(total_iteration):
+        for dir in range(num_directions):
+            # print('yielding')
+            yield (data[idx, groups[:, dir]].flatten(), bvals[:, groups[:, dir]].ravel(), b0_indices[:, groups[:, dir]].ravel())
+
+def osipi_fit(fitfunc, data_bvals):
+    data, bvals, b0_indices = data_bvals
```
oliverchampion, line 36: Does it make more sense to use dictionaries for these?

utilities/process/diffusion_utils.py
```diff
@@ -0,0 +1,202 @@
...
+    # invert the sort to return the shell indices of the unsorted bvalues
+    bvals_inverse_sort = np.empty_like(bvals_indices)
+    bvals_inverse_sort[bvals_indices] = np.arange(bvals_indices.size)
+
+    # find the mean b-values of the shells
+    shells = [np.mean(bvals[index==idx]) for idx in set(index)]
+    indices = index[bvals_inverse_sort]
+    return shells, indices, indices<atol #, bvals_indices
+
+def find_directions(bvecs_raw, zero_indices, atol=1, rotationally_symmetric=True):
```
oliverchampion, line 73: Looks good. Are we doing a lot with vectors?

oliverchampion (commented):
I did my best reviewing. Not 100% sure what everything does, but it seems good to me. good that you have added unit testing too :).

IvanARashid:
Thanks, I'll attempt to implement a version of this in the wrapper

etpeterson:
> Thanks, I'll attempt to implement a version of this in the wrapper

I wasn't trying to get you to work on it, just throwing this out as an initial implementation. I do think adding it to the wrapper is the best approach though.

I was thinking of putting it in the wrapper like you did the multi voxel fit. Probably each algorithm should implement their own data generator though. I need to think about the structure a little more.

WrapImage/nifti_wrapper.py
etpeterson: Yes, I think Ivan has a PR for that so maybe there's a conflict between these. I guess either his or mine should get merged first and then the other will need to update the other.

utilities/process/diffusion_utils.py
etpeterson: Not really right now. I was thinking some algorithms will want to operate on a single direction at a time so that's why I wrote this, to be able to automatically split those up.

WrapImage/nifti_wrapper.py
etpeterson: The source could come from anywhere, this splits the data up based on the bvals and directions. The b0 indices are included in order to give the b0 data to every fit.

True, I should update it with some more documentation.

etpeterson:
Based on our discussion I'll move this over to the base wrapper.

PR #84, opened 2024-10-08
Developed NIFTI 4d viewer with voxel intensity plots.

AhmedBasem20:
Added a new section to the TF2.4 website (the final piece of GSoC):
image

## Features
- **Multi-viewport Visualization** Users can upload a NIFTI file, which is displayed in three synchronized viewports (axial, sagittal, and coronal) using the CornerstoneJS suite.
- **Voxel Intensity Plot** A voxel intensity plot will be generated and displayed when you drag or scroll through an image viewport.
- For now, the intensity values are on the Y axis, and the total time points on the X axis.

## Testing

I used software like ITK-SNAP and MRIcroGL to test the accuracy of the cursor position and the intensity plots and I got accurate results.

image

## TODO

- Update the cursor position and the intensity plot on mouse click event (for now it is updated on scroll and mouse drag only)
- Create a pipeline to generate NIFTI files from the repository and display it initially on the viewer, thoughts?
- Improve the intensity plot by adding corresponding b-values on the X axis and the fitted curve from d, dp, and f images.

## Demo:
https://ahmedbasem20.github.io/TF2.4_IVIM-MRI_CodeCollection/nifti-viewer/

### DIFF
--- a/website/nifti-viewer/index.js
+++ b/website/nifti-viewer/index.js
@@ -89,9 +89,9 @@ function handleVoxelClick(currentVoxel) {
     const [nx, ny, nz, nt] = dims;
     let [voxelX, voxelY, voxelZ] = currentVoxel;
 
-    voxelX = Math.min(Math.max(Math.round(voxelX), 1), nx) - 1;
-    voxelY = Math.min(Math.max(ny - Math.round(voxelY), 1), ny) - 1;
-    voxelZ = Math.min(Math.max(nz - Math.round(voxelZ), 1), nz) - 1;
+    voxelX = Math.min(Math.max(Math.round(voxelX), 1), nx);
+    voxelY = Math.min(Math.max(ny - Math.round(voxelY), 1), ny);
+    voxelZ = Math.min(Math.max(nz - Math.round(voxelZ), 1), nz);
 
     const voxelValues = getVoxelValuesAcrossTime(voxelX, voxelY, voxelZ, nx, ny, nz, nt);
     updateVoxelCoordinatesDisplay(voxelX + 1, voxelY + 1, voxelZ + 1);


website/nifti-viewer/index.js
```diff
@@ -0,0 +1,232 @@
...
+
+// Enable tools and interactions for the displayed image
+function setupImageTools(element, imageIdObject) {
+    const numberOfSlices = cornerstone.metaData.get('multiFrameModule', imageIdObject.url).numberOfFrames;
+    const stack = {
+        currentImageIdIndex: imageIdObject.slice.index,
+        imageIds: Array.from({ length: numberOfSlices }, (_, i) => `nifti:${imageIdObject.filePath}#${imageIdObject.slice.dimension}-${i},t-0`)
+    };
+
+    cornerstoneTools.addStackStateManager(element, ['stack']);
```
etpeterson, line 75: I think we also want a rescale tool, sometimes I can barely see anything in those images.

website/nifti-viewer/index.js
```diff
@@ -0,0 +1,232 @@
...
+    voxelY = Math.min(Math.max(ny - Math.round(voxelY), 1), ny) - 1;
+    voxelZ = Math.min(Math.max(nz - Math.round(voxelZ), 1), nz) - 1;
+
+    const voxelValues = getVoxelValuesAcrossTime(voxelX, voxelY, voxelZ, nx, ny, nz, nt);
+    updateVoxelCoordinatesDisplay(voxelX + 1, voxelY + 1, voxelZ + 1);
+    plotVoxelData(voxelValues);
+}
+
+// Extract voxel values across all time points
+function getVoxelValuesAcrossTime(x, y, z, nx, ny, nz, nt) {
```
etpeterson, line 102: This doesn't seem to be correct, select bright parts of images but the value is low. I select dark parts of the image but the value is high. Most likely x, y, and z aren't what you think they are. But also is there no cornerstone based way to do this?

etpeterson (changes requested):
Nicely interactive but I found a few bugs.

website/nifti-viewer/index.js
```diff
@@ -0,0 +1,232 @@
...
+
+// Load NIfTI file and display the axial, sagittal, and coronal views
+function loadAllFileViews(file) {
+    const fileURL = URL.createObjectURL(file);
+    const imageId = `nifti:${fileURL}`;
+
+    cornerstoneNIFTIImageLoader.nifti.loadHeader(imageId).then((header) => {
+        dims = [...header.voxelLength, header.timeSlices];
+        loadAndViewImage(document.getElementById('nifti-image-z'), `${imageId}#z,t-0`, 'axial');
+        loadAndViewImage(document.getElementById('nifti-image-x'), `${imageId}#x,t-0`, 'sagittal');
```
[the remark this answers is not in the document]
AhmedBasem20: There is a bug in synchronizing the sagittal view initially that I'm still trying to figure out.
For now you can fix this by scrolling through the view that is not synched, it will work normally after that.

website/nifti-viewer/index.js
AhmedBasem20: You mean a tool for zooming in/out, right?

website/nifti-viewer/index.js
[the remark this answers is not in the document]
etpeterson: Actually I'm noticing it's both the sagittal and coronal views that don't act correctly.

website/nifti-viewer/index.js
etpeterson: It's still not working. See how I have a very bright part selected but the values are quite low. Other areas show up much higher on the plot than this.
I'm guessing it's either image orientation differences or still some indexing problems. I don't see how you're handling the orientation differences, so that's one place to start. What I mean by that is that indexing into the raw data doesn't necessarily equate to the displayed x, y, z. There's a possible axis flip and permutation based on the raw data so it's always displayed in the correct orientation.
image

website/nifti-viewer/index.js
etpeterson: No, brightness/contrast is what they seem to call it here. https://www.cornerstonejs.org/live-examples/videocolor

website/nifti-viewer/index.js
AhmedBasem20: There is orientation difference and handled it here: https://github.com/OSIPI/TF2.4_IVIM-MRI_CodeCollection/blob/738b6f1aab7769f5d0135e266d4b2e64df4d0edc/website/nifti-viewer/index.js#L92-L94
I used the complement of y and z due to the axis flip. and the indexing seems accurate according to the images I tested. Can you share the image you're using here so that I can reproduce the issue?

website/nifti-viewer/index.js
AhmedBasem20: got it, will work on that.

website/nifti-viewer/index.js
etpeterson: I can't share the images because of patient privacy, but you can't count on that transformation always being the same. I opened the same image in yours and a full viewer and I see the images and notations are flipped left/right. Here's a screenshot showing that and some header info.
image

website/nifti-viewer/index.js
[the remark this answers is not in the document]
etpeterson: Same problem sharing images, sorry. Now I'm not seeing the coronal view misbehave. Definitely the sagittal one though. Sounds like you see that too though.

etpeterson:
Hey @AhmedBasem20 are you going to be able to get more work done on this?

etpeterson:
> Hi @etpeterson Sorry I've been very busy recently but hopefully I'll get more work in and address your comments next week.

Any chance this'll get done?

PR #96, opened 2025-03-20
Implement checks for fulfilment of algorithm requirements in the OsipiBase class

### LINKED ISSUE #45: [Feature] Implement checks for fulfilment of algorithm requirements in the OsipiBase class
### Feature description

Currently, the OsipiBase class is used as a parent class for the standardisation of the submitted algorithms. It is also used as a common interface which we use in our testing, allowing for mass-calling of function fitting.

However, one thing we would like to implement in order to make things more robust is to check the inputs and compare them to the requirements of the submitted code. This is in order to raise relevant errors when we know that something will not work due to the input of e.g. init of OsipiBase, or OsipiBase.osipi_fit() not being compatible with the requirements of the fitting algorithm.

### Describe the solution

The stadardised implementations of the algorithms can have attributes which describe the requirements of the inputs. E.g. specification of function bounds, number of b-values, number of b-value thresholds, etc (see the required_* attributes in src/standardized/IAR_LU_biexp.py. Write functions that makes sure that the inputs to OsipiBase follow these requirements. Call these checks in the OsipiBase.osipi_fit() function.

### Describe alternatives

_No response_

### Additional context

IVIM fitting algorithms can be categorised into e.g. least-squares fits, segmented least-squares fits, Bayesian, variable projection, etc. These have different types of requirements. The segmented algorithms require some kind of segmentation of the b-scale, i.e. a b-value threshold. Therefore, these algorithms would need this parameter specified. However, it could also be that the specific algorithm in question already has some default value, making it a non-requirement in the case of actually initiating the OsipiBase class. So to understand the requirements of each submission, one would need to actually dig into the code. However, for the implementation of this feature, it is not necessarily needed. You essentially only need to write code that compares inputs with the requirements specified in the standardised implementation of the algorithm.

### Are you working on this?

None

harshithasudhakar:

Implemented a centralized validation logic for inputs in the base class.

Fixes #45 

### Checklist

- [ ] Update Docs & Guides


.gitignore
```diff
@@ -25,4 +25,6 @@ nosetests.xml
 coverage.xml
 *.pyc
 phantoms/MR_XCAT_qMRI/*.json
-phantoms/MR_XCAT_qMRI/*.txt
\ No newline at end of file
+phantoms/MR_XCAT_qMRI/*.txt
+osipivenv/
```
etpeterson, line 29: I don't think this is needed? It's probably local to your machine.

src/standardized wip/IAR_LU_linear.py
```diff
@@ -39,6 +39,10 @@ def __init__(self, bvalues=None, thresholds=None, bounds=None, initial_guess=Non
         super(IAR_LU_linear, self).__init__(bvalues, thresholds, bounds, initial_guess)
         
         # Check the inputs
+        self.osipi_check_required_bvalues()
```
etpeterson, line 42: I like where you're going with this, but adding these checks everywhere is cumbersome. I'm not sure exacly the best approach, but a few come to mind.
1. Adding them to the base class. This would abstract it away, but also make it less flexible when the values are set.
2. Creating setters for these values, and in the setters we can call these methods.

I'm open to other suggestions too. Maybe @IvanARashid has some ideas here?

etpeterson:
Thanks for this. It's a good start. I'd like some feedback from @IvanARashid on this too. Basically I have one main suggestion for making it easier to maintain. It would also be great to get a test that goes through the algorithms and somehow verifies the validators are running.

etpeterson (changes requested):

etpeterson:
Also looks like tests are failing because of this. I think it's too strict, or not all the algorithms are acting in quite the standard way.

harshithasudhakar:
Sorry for the delay, I've made the changes now. Please let me know if there are any more changes I should make.

src/standardized/ETP_SRI_LinearFitting.py
```diff
@@ -41,7 +41,7 @@ def __init__(self, bvalues=None, thresholds=None, bounds=None, initial_guess=Non
             Our OsipiBase object could contain functions that compare the inputs with
             the requirements.
         """
-
+        super(ETP_SRI_LinearFitting, self).__init__(bvalues=bvalues, thresholds=thresholds, bounds=bounds, initial_guess=initial_guess)
```
etpeterson, line 44: Looks like nearly duplicated `super` calls.

src/wrappers/OsipiBase.py
```diff
@@ -16,6 +16,10 @@ def __init__(self, bvalues=None, thresholds=None, bounds=None, initial_guess=Non
         self.initial_guess = np.asarray(initial_guess) if initial_guess is not None else None
         self.use_bounds = True
         self.use_initial_guess = True
+
+        # Validate the inputs
+        self.osipi_validate_inputs()
```
etpeterson, line 21: I think bvalues are not required here, and I believe that will fail the check. Should we require it at this level? I hate to add too many requirements so perhaps we need 2 levels of checking. A lenient and a strict where the lenient passes `None` but strict does not. Then the check here can be lenient but the check before fitting can be strict.

src/wrappers/OsipiBase.py
```diff
@@ -53,7 +57,8 @@ def osipi_fit(self, data, bvalues=None, **kwargs):
         """Fits the data with the bvalues
         Returns [S0, f, Dstar, D]
         """
-        
+        self.osipi_validate_inputs()
```
etpeterson, line 60: This is double checking things. Maybe we just need the check here and not even the lenient and strict I was talking about above? That would be re-checking for each fit but perhaps that's not too much overhead? Or we could try to not re-check if they've been checked already?

etpeterson (changes requested):

etpeterson:
Still failing tests.

src/wrappers/OsipiBase.py
IvanARashid: Yeah I think this is the right spot, as this is where we pass most of the arguments we want to check. If we do it before the loop over the voxels, it will only be once per call and not per voxel, so I think it will be fine.

src/standardized/ETP_SRI_LinearFitting.py
IvanARashid: Yeah I think we have it like this for the other algorithms, but only one of them should be kept.

src/wrappers/OsipiBase.py
harshithasudhakar: Sure, I'll change it to a lenient check in init, and a strict check in fit

src/wrappers/OsipiBase.py
```diff
@@ -201,69 +215,63 @@ def osipi_print_requirements(self):
...
         """Query if the selection dimension is fittable"""
         
-        #accepted = self.accepted_dimensions()
-        #if dim < 0 or dim > len(accepted):
-            #return False
-        #return accepted[dim]
+        min_dim, max_dim = self.osipi_accepted_dimensions()
+        if not (min_dim <= dim <= max_dim):
+            raise ValueError(f"Dimension {dim}D not supported. Requires {min_dim}-{max_dim}D spatial data")
         return True
```
etpeterson, line 241: I don't think we want to return anything for any of these anymore.

etpeterson (changes requested):
Looking pretty good but the tests are still failing as well.

src/wrappers/OsipiBase.py
harshithasudhakar: I've removed them now, it should pass the tests now.

etpeterson (changes requested):
The tests are still failing because of the checks you added. For example here: https://github.com/OSIPI/TF2.4_IVIM-MRI_CodeCollection/actions/runs/14249360771/job/39959244676#step:6:79
Either the checks need to be modified or how the algorithms are called. Probably the algorithm calls aren't consistent or as they should be, but I haven't looked into it. Let me know if you need some help with this.

etpeterson:
Hi @harshithasudhakar, thanks for this. I think the major remaining issue is the failing tests. Are you seeing these? 
image

etpeterson (changes requested):
The failing tests need to be resolved before merging.

harshithasudhakar:
> Hi @harshithasudhakar, thanks for this. I think the major remaining issue is the failing tests. Are you seeing these? image

I've fixed them, the test cases should pass now.

etpeterson (changes requested):
Tests are still failing, sorry.

etpeterson:
Just checking on this, are you planning on finishing this up?

src/wrappers/OsipiBase.py
```diff
@@ -201,68 +217,60 @@ def osipi_print_requirements(self):
...
     def osipi_check_required_thresholds(self):
         """Checks if the number of input thresholds fulfil the algorithm requirements"""
-        
-        #if (len(self.thresholds) < self.required_thresholds[0]) or (len(self.thresholds) > self.required_thresholds[1]):
-            #print("Conformance error: Number of thresholds.")
-            #return False
-        #else: 
-            #return True
-        return True
+        if not hasattr(self, "required_thresholds"):
```
oliverchampion, line 258: I am wondering whether for some attributes, like "required_thresholds", we should add "default" values when tests fail. Like, 200 would be an okay value for most body applications. Then we can do with a warning stating that the default value is used, instead.

src/wrappers/OsipiBase.py
```diff
@@ -201,68 +217,60 @@ def osipi_print_requirements(self):
...
         
     def osipi_check_required_bounds(self):
         """Checks if input bounds fulfil the algorithm requirements"""
-        #if self.required_bounds is True and self.bounds is None:
-            #print("Conformance error: Bounds.")
-            #return False
-        #else:
-            #return True
-        return True
+        if self.required_bounds is False and self.bounds is None:
```
oliverchampion, line 267: Same for bounds. I think if no bounds are given, but bounds will be used, then we should give some default bounds.

src/wrappers/OsipiBase.py
```diff
@@ -201,68 +217,60 @@ def osipi_print_requirements(self):
...
-            #return False
-        #else:
-            #return True
-        return True
-
-    
-    def osipi_check_required_bvalues():
-        """Minimum number of b-values required"""
-        pass
+        if self.required_initial_guess is False and self.initial_guess is None:
```
oliverchampion, line 272: same for initial guess :)

src/wrappers/OsipiBase.py
```diff
@@ -201,68 +217,60 @@ def osipi_print_requirements(self):
...
+        """Check if the spatial dimensions (excluding b-values) are supported."""
+        spatial_dim = dim - 1  # Exclude last axis (b-values)
+        min_dim, max_dim = self.osipi_accepted_dimensions()
+        if not (min_dim <= spatial_dim <= max_dim):
+            raise ValueError(
+                f"Spatial dimensions {spatial_dim}D not supported. "
+                f"Requires {min_dim}-{max_dim}D."
+            )
     
     def osipi_check_required_bvalues(self):
```
oliverchampion, line 247: b-values is another story. Here we cannot give a "default"

oliverchampion (commented):
Hey! I was thinking it may make more sense to come up with some "default" values for initial guess, bounds and cut-off when they are missing, rather then give an error.

Now, I personally solve the missing of these values in my wrapper. But it may be better to have the Osipi base to come up with default values that are identical for all algorithms.

oliverchampion:
what is the status of this PR? should we merge/close?

oliverchampion:
Not sure if all of this is still equally relevant, but having tests for default values is relevant. 
@IvanARashid, I think you have most insight in the wrapper. Can you check whether all these tests are still relevant, also after the recent push by devguru?

oliverchampion:
@ericpete @harshithasudhakar , sorry for late reactions. I'm more active now :). 
Is this PR still relevant? If yes, please resolve the conflicts and previous comment. If not relevant, I will close it over the next weeks :)

PR #99, opened 2025-04-20
Added the feature to read DICOM images into a 4D Nifti image 

jph6366:
### Allow reading of DICOM images

Spent a few days getting acclimated to DICOM concepts (shoutout [DicomIsEasy](https://dicomiseasy.blogspot.com/p/introduction-to-dicom.html)) and investigated the best options for this feature request.

I initially wanted to use [ITK](https://itk.org/) like it is [here](https://github.com/PaulHax/knee-sarg/blob/main/knee_sarg/ingest/ingest_dicom.py), but it seems perhaps a configure-first philosophy might suit this codebase by introducing a new feature with less complexity added to future development. I opted to use [dcm2nii](https://www.nitrc.org/plugins/mwiki/index.php/dcm2nii:MainPage), which converts diffusion vectors andit is written in C, so it is very portable, and It attempts to preserve diffusion gradient information. [dcm2nii](http://www.mccauslandcenter.sc.edu/mricro/mricron/dcm2nii.html) is the predecessor to dcm2niix. It is deprecated for modern images, but does handle image formats that predate DICOM.

dicom2niix documentation is good and the repository is open source, also they list many alternatives.

Fixes _#68_

### Checklist

- [ ] Test nifti_wrapper.py CLI
- [ ] Updated Dockerfile to install dicom2niix latest release
- [ ] Update README


WrapImage/nifti_wrapper.py
```diff
@@ -1,11 +1,51 @@
...
+    try:
+        res = subprocess.run(
+            [
+                "dcm2niix",
+                "-f", "%s_%p", # dcm2niix attempts to provide a sensible file naming scheme
+                "-o", vol_dir,
+                "-z", "y", #specifying compressed nii.gz file
+                # https://www.nitrc.org/plugins/mwiki/index.php/dcm2nii:MainPage
+                # for further configuration for general usage see page above
+                vol_dir
```
etpeterson, line 25: Is this outputting into the same folder as the input? It should probably be a temporary folder or perhaps sub folder to prevent mixing of the image types.

WrapImage/nifti_wrapper.py
```diff
@@ -1,11 +1,51 @@
...
+        if len(nifti_files) > 1:
+            raise RuntimeError("Multiple NIfTI files generated; expected one 4D NIfTI.")
+
+        bval_files = list(vol_dir.glob("*.bval"))
+        bvec_files = list(vol_dir.glob("*.bvec"))
+        bval_path = str(bval_files[0]) if bval_files else None
+        bvec_path = str(bvec_files[0]) if bvec_files else None
+
+
+        return vol_dir / "dicom2nifti.nii.gz", bval_path, bvec_path
```
etpeterson, line 44: It looks like the nifti name is hard coded. Shouldn't it be specifically the file that's written this run? For example, if this is run repeatedly what happens?

WrapImage/nifti_wrapper.py
```diff
@@ -1,11 +1,51 @@
...
+        nifti_files = list(Path(vol_dir).glob("*.nii.gz"))
+        if not nifti_files:
+            raise RuntimeError("No NIfTI file generated by dcm2niix.")
+        if len(nifti_files) > 1:
+            raise RuntimeError("Multiple NIfTI files generated; expected one 4D NIfTI.")
+
+        bval_files = list(vol_dir.glob("*.bval"))
+        bvec_files = list(vol_dir.glob("*.bvec"))
+        bval_path = str(bval_files[0]) if bval_files else None
+        bvec_path = str(bvec_files[0]) if bvec_files else None
```
etpeterson, line 41: This is all making the assumption that we only want one. What if they're different, which do we choose then? I'm not exactly sure what dcm2niix outputs, but probably best is to require just one output and error if there are more.

Docker/Dockerfile
```diff
@@ -5,6 +5,13 @@ WORKDIR /usr/src/app
 RUN apt-get update && apt-get install -y --no-install-recommends \
     build-essential \
     libssl-dev \
+    wget \
+    pigz \
+    && wget https://github.com/rordenlab/dcm2niix/archive/refs/tags/v1.0.20241211.tar.gz \
```
etpeterson, line 10: This could be a separate build stage in the dockerfile. That simplifies the cleanup.

I'm also wondering about the version. Is there a way we can select the latest stable?

etpeterson (changes requested):
Good start! I have a few comments in specific places.

WrapImage/nifti_wrapper.py
jph6366: For handling batch conversion for many files we could leverage the yaml config for batch processing, as [specfied in the dicom2niix repo](https://github.com/rordenlab/dcm2niix/blob/master/BATCH.md), so we can provide that option and updated instructions in the README.

Docker/Dockerfile
jph6366: it seems the easiest way to do it is just pin the latest version as so. 

We could query the GitHub API but then our docker environment will get rate limited without a personal access token for authorization. Ideally, we dont want to increase our dependency and just lean towards manual updates based on alerting users to make an issue to update the dockerfile.

there is an [update check option](https://www.nitrc.org/plugins/mwiki/index.php/dcm2nii:MainPage#:~:text=u%20Update%20check%3A%20(%22%2Du%22%2C%20UNIX%20only).%20This%20command%20will%20cause%20the%20software%20report%20if%20there%20are%20any%20newer%20versions%20of%20dcm2niix%20available%20online.%20Note%2C%20this%20command%20only%20checks%20for%20updates%2C%20it%20does%20not%20update%20the%20software.) to add to our dicom2niix subprocess call, this command will cause the software report if there are any newer versions of dcm2niix available online. (UNIX ONLY)

WrapImage/nifti_wrapper.py
jph6366: > I'm not exactly sure what dcm2niix outputs, but probably best is to require just one output and error if there are more.

dicom2niix handles it nicely really by just adding characters in an alphabetical order

e.g. 

len(nifti_files) = 0  *.bval, *.bvec *.nii.gz  

len(nifti_files) = 1  *a.bval, *a.bvec *a.nii.gz  

....

len(nifti_files) = 2  *c.bval, *c.bvec *c.nii.gz

jph6366:
I'm going to refactor in order to...

- address general fixes and comments/threads
- incoporate the batch processing features by automating the generation of the `batch_config.yaml` given file inputs
- add an `nifti_wrapper` CLI option to specify whether dicom conversion is needed at all and increase seamless handling of file input
- Integrate best practices for multi-stage docker build and more, then try reduce the build time without compromising features

WrapImage/inquire_dicom2niix.py
```diff
@@ -0,0 +1,233 @@
...
+        if not nifti_files:
+            raise RuntimeError("No NIfTI (.nii.gz) files were generated.")
+
+        bval_files = list(out_dir.glob("*.bval"))
+        bvec_files = list(out_dir.glob("*.bvec"))
+        bval_path = str(bval_files[0]) if bval_files else None
+        bvec_path = str(bvec_files[0]) if bvec_files else None
+
+
+        return nifti_files[0], bval_path, bvec_path
```
etpeterson, line 139: As previously, there doesn't seem to be any error handling in the case of multiple nifti files. It looks like this converts everything and then forgets all but the first one. We either want to error if more than one are present or process them all sequentially.

WrapImage/inquire_dicom2niix.py
```diff
@@ -0,0 +1,233 @@
...
+            check=True
+        )
+        
+        nifti_files = list(Path(out_dir).glob("*.nii.gz"))
+        if not nifti_files:
+            raise RuntimeError("No NIfTI (.nii.gz) files were generated.")
+
+        bval_files = list(out_dir.glob("*.bval"))
+        bvec_files = list(out_dir.glob("*.bvec"))
+        bval_path = str(bval_files[0]) if bval_files else None
```
etpeterson, line 135: The bval and bvec are required.

Docker/dicom2nifti/Dockerfile
```diff
@@ -0,0 +1,41 @@
+FROM ubuntu:jammy AS build
```
etpeterson, line 1: Could we base this off a slim of some sort?

Docker/dicom2nifti/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    build-essential \
+    cmake \
+    git \
+    libssl-dev \
+    wget \
+    pigz \
+    ca-certificates \
+    && update-ca-certificates \
+    && wget https://github.com/rordenlab/dcm2niix/archive/refs/tags/v1.0.20241211.tar.gz -O /tmp/dcm2niix.tar.gz \
```
etpeterson, line 12: If we're going to fix a version let's at least make it a variable.

Docker/dicom2nifti/Dockerfile
etpeterson, line 1: I think the other docker image has a build test, can we build test this one too?

etpeterson (changes requested):
Thanks for keeping at this, it's converging. Several more comments though.

WrapImage/inquire_dicom2niix.py
```diff
@@ -0,0 +1,233 @@
...
+        except RuntimeError as err:
+            print(f"❌ Conversion failed: {err}")
+
+    else:
+        print("📥📥📥 Multiple inputs detected — generating batch config and using `dcm2niibatch`...")
+        config_path = "batch_config.yaml"
+        generate_batch_config(input_dirs, output_dirs, config_path=config_path)
+
+        try:
+            with subprocess.Popen(
```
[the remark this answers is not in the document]
jph6366: Yes it will be removed, it is not presently maintained either compared the dcm2niix command.

WrapImage/inquire_dicom2niix.py
jph6366: So after testing it, it will actually handle two differing sets of dicom images in the same directory gracefully. It will properly create separate nifti files as well with respect to bval bvec files accompanied with BIDS-compatible JSON files. 

To have more control over the quantity of the output, there is Single file mode: ("-s y" or "-s n") and option to only convert this series Number: (e.g. "-n 12"). By default it will identify images that are in the same Series and group them into a nii.gz file (w.r.t bval and bvec). Lastly there is the option to merge 2D slices: ("-m y" or "-m n"), where images from the same series are stacked into a single NIfTI image regardless of study time, echo, coil, orientation, etc.

e.g. (BY DEFAULT) If there are 3 dcm images all of the sames series, there should be only 1 nifti file as specified, and if there are 5 dcm images with values representing 3 different series, then there should be 3 nifti files.

pyproject.toml
```diff
@@ -0,0 +1,46 @@
+[project]
+name = "TF2.4_IVIM-MRI-CodeCollection"
+version = "0.1.0"
+description = "IVIM MRI signal simulation and DICOM/NIfTI processing tools"
+authors = [{ name="Jackson Hardee", email="jphardee@gmail.com" }]
+requires-python = ">=3.11"
```
etpeterson, line 6: Minimum python officially supported is 3.9

pyproject.toml
```diff
@@ -0,0 +1,46 @@
+[project]
+name = "TF2.4_IVIM-MRI-CodeCollection"
+version = "0.1.0"
+description = "IVIM MRI signal simulation and DICOM/NIfTI processing tools"
+authors = [{ name="Jackson Hardee", email="jphardee@gmail.com" }]
```
etpeterson, line 5: Should probably be OSIPI? Might be good to get someone else's feedback on this too.

pyproject.toml
```diff
@@ -0,0 +1,46 @@
+[project]
+name = "TF2.4_IVIM-MRI-CodeCollection"
+version = "0.1.0"
+description = "IVIM MRI signal simulation and DICOM/NIfTI processing tools"
```
etpeterson, line 4: IVIM signal processing repository

pyproject.toml
```diff
@@ -0,0 +1,46 @@
...
+authors = [{ name="Jackson Hardee", email="jphardee@gmail.com" }]
+requires-python = ">=3.11"
+readme = "README.md"
+license = {text = "MIT"}
```
etpeterson, line 8: Apache, not MIT

Docker/dicom2nifti/DICOM2NiFTI.png
etpeterson, line 1: I'm not sure what this supposed to be showing. This is what I'm seeing in github at least. If we need this, is it something we could link out to rather than having in the repo? 
image

Docker/dicom2nifti/Dockerfile
etpeterson: It still doesn't look like it's tested. See the docker-build-and-run.yml test.

etpeterson (changes requested):

Docker/dicom2nifti/DICOM2NiFTI.png
jph6366: we do not need this. this is showcasing the spatial orientation of generated ivim signal of a 2D slice (from a dicom image converted nii.gz) compared to 3D Volume (from a nii.gz).

jph6366: image
views like this

etpeterson: > we do not need this. this is showcasing the spatial orientation of generated ivim signal of a 2D slice (from a dicom image converted nii.gz) compared to 3D Volume (from a nii.gz).

Ok, let's remove it then. That kind of information can be found elsewhere.

.github/workflows/docker-build-and-run.yml
```diff
@@ -76,3 +76,73 @@ jobs:
...
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build Docker image
+        run: |
+          docker build -t tf2.4_ivim-mri_codecollection -f Docker/dicom2nifti/Dockerfile .
+
+      - name: Run Docker container
+        run: |
+          docker run --rm --name TF2.4_IVIM-MRI_CodeCollection \
+            -v ${{ github.workspace }}:/usr/src/app \
```
etpeterson, line 119: Why mount two folders to the same place? That seems a little odd.

etpeterson:
The new test fails: https://github.com/OSIPI/TF2.4_IVIM-MRI_CodeCollection/actions/runs/14710507373/job/41395983046?pr=99

etpeterson (changes requested):

etpeterson:
Getting close. Just smaller comments now.

etpeterson:
Sorry, looks like the test still fails.

etpeterson (changes requested):
Just because the test fails

.github/workflows/docker-build-and-run.yml
```diff
@@ -98,9 +98,10 @@ jobs:
 
       - name: Verify input files
         run: |
+          IVIM_SIM_DIR = "${{ github.workspace }}/ivim_simulation"
```
etpeterson, line 101: This test fails. I think because of the whitespace here.

etpeterson (changes requested):
Sorry for the long delay. I was at a conference. Still just that test failure.

.github/workflows/docker-build-and-run.yml
```diff
@@ -98,8 +98,8 @@ jobs:
 
       - name: Verify input files
         run: |
-          IVIM_SIM_DIR = "${{ github.workspace }}/ivim_simulation"
-          for file in ivim_simulation_0000.dcm, ivim_simulation_0001.dcm, ivim_simulation_0002.dcm ivim_simulation_0003.dcm ivim_simulation_0004.dcm; do
+          IVIM_SIM_DIR="${{ github.workspace }}/ivim_simulation"
```
etpeterson, line 101: Still failing. https://github.com/OSIPI/TF2.4_IVIM-MRI_CodeCollection/actions/runs/15230335594/job/42931601286?pr=99
Probably not quite the right path, especially the comma at the end.
You should be able to test this on your end by running the actions rather than waiting for me to run and check the tests.

etpeterson (changes requested):

pyproject.toml
```diff
@@ -0,0 +1,44 @@
+[project]
+name = "TF2.4_IVIM-MRI-CodeCollection"
+version = "0.1.0"
+description = "IVIM MRI signal processing and NIfTI/DICOM processing tools"
+requires-python = ">=3.9"
+readme = "README.md"
+dependencies = [
```
etpeterson, line 7: I'm noticing these dependencies don't completely match with those in requirements.txt. Is that intentional?

etpeterson (changes requested):
Tests are passing! Just took another look across the code and came up with one final question.

PR #101, opened 2025-05-22
Added a segmented-fitting algorithm based on the recommendation paper

Bneijndorff:
A segmented fitting algorithm based on the recommendation IVIM paper was added.

The algorithm first fits D using a WLS mono-exponential fit using data of all bvalues above the cutoff value
Then f is estimated from the intercept with b=0 from the mono-exponential fit and the measured signal for b=0
Last D* is estimated from after fixing D and f and fitting the bi-exponential IVIM formula.

In the way it's now coded only bounds for D* are implemented and the initial guess for D* = 10*D

I did do some simple tests and the algorithm seems to work.

### Checklist

- [ ] Update Docs & Guides

### DIFF
--- a/phantoms/MR_XCAT_qMRI/sim_ivim_sig.py
+++ b/phantoms/MR_XCAT_qMRI/sim_ivim_sig.py
@@ -4,6 +4,10 @@ import nibabel as nib
 import json
 import argparse
 import os
+
+import sys
+sys.path.append('C:/TF_IVIM_OSIPI/TF2.4_IVIM-MRI_CodeCollection')
+
 from utilities.data_simulation.Download_data import download_data
 
 ##########
@@ -19,7 +23,7 @@ def phantom(bvalue, noise, TR=3000, TE=40, motion=False, rician=False, interleav
         states = [1]
     for state in states:
         # Load the .mat file
-        mat_data = loadmat('../../download/Phantoms/XCAT_MAT_RESP/XCAT5D_RP_' + str(state) + '_CP_1.mat')
+        mat_data = loadmat('download/Phantoms/XCAT_MAT_RESP/XCAT5D_RP_' + str(state) + '_CP_1.mat')
 
         # Access the variables in the loaded .mat file
         XCAT = mat_data['IMG']
@@ -429,7 +433,7 @@ if __name__ == '__main__':
     elif args.bvalue:
         bvalues = {"cmd": args.bvalue}
     else:
-        bvalues = parse_bvalues_file("b_values.json") 
+        bvalues = parse_bvalues_file("phantoms\\MR_XCAT_qMRI\\b_values.json")
 
     
     noise = args.noise
--- a/src/original/TF_reference/segmented_IVIMfit.py
+++ b/src/original/TF_reference/segmented_IVIMfit.py
@@ -1,0 +1,76 @@
+"""
+Code to calculate IVIM maps
+
+Segmented fitting approach
+First fit D using a WLLS
+
+"""
+
+import numpy as np
+import statsmodels.api as sm
+import scipy
+
+def ivim_biexp(bvalues, D, f, Dp, S0=1):
+    return (S0 * (f * np.exp(-bvalues * Dp) + (1 - f) * np.exp(-bvalues * D)))
+
+
+def segmented_IVIM_fit(bvalues, dw_data, b_cutoff = 200, bounds=([0.0001, 0.0, 0.001], [0.004, 0.7, 0.01])):
+    """
+        A segmented fitting implementation for a bi-exponential model.
+        First D is fitted using a mono exponential model on all signal above the bvalue cutoff using an iterative WLLS
+        Then f is fitted by using the b=0 intercept from the mono expontential fit and substracting this from the measured signal at b=0
+        Then D* is fitted using a bi-exponential model with fixed D and f
+
+        """
+    bvalues_D = bvalues[bvalues >= b_cutoff]
+    dw_data_D = dw_data[bvalues >= b_cutoff]
+
+    D, b0_intercept = d_fit_iterative_wls(bvalues_D, np.log(dw_data_D))
+
+    D = D
+    S0 = dw_data[bvalues == 0].item()
+    f = (S0 - b0_intercept) / S0
+
+    Dp = scipy.optimize.least_squares(
+    fun=lambda Dp: ivim_biexp(bvalues, D, f, Dp[0]) - dw_data,
+    x0=D*10,  # Initial guess for D*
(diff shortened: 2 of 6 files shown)


phantoms/MR_XCAT_qMRI/sim_ivim_sig.py
```diff
@@ -4,6 +4,10 @@
 import json
 import argparse
 import os
+
+import sys
+sys.path.append('C:/TF_IVIM_OSIPI/TF2.4_IVIM-MRI_CodeCollection')
```
etpeterson, line 9: This is your local path but we want to avoid this in the repository because we want to keep it general. It's guaranteed to fail for anyone not on windows, for example.

etpeterson (changes requested):
It's not immediately obvious what's causing the test failures. It seems to be running but just failing the tests. Does it work for you locally?

oliverchampion:
Possibly fixing the nans and inf problem will solve the svd problem too. Otherwise I'm happy to take a look!

Bneijndorff:
Clipping the signal to a minimum of 0.01 fixed the issues and now it passes all tests.

src/original/TF_reference/segmented_IVIMfit.py
```diff
@@ -0,0 +1,78 @@
...
+        signal_pred = results.predict(bvals)
+        residuals = log_signal - signal_pred
+        new_weights = 1 / np.maximum(residuals ** 2, 1e-10)
+
+        if np.linalg.norm(new_weights - weights) < tolerance:
+            break
+        weights = new_weights
+
+    ln_b0_intercept, D_neg  = results.params
+    b0_intercept = np.exp(ln_b0_intercept)
```
oliverchampion, line 76: I'm not familiar with the WLLS, but it seems reasonable. I assume you found this somewhere :)?

src/original/TF_reference/segmented_IVIMfit.py
```diff
@@ -0,0 +1,78 @@
...
+
+    tolerance: the tolerance for early stopping of the WLS
+    """
+
+    bvals = sm.add_constant(bvalues_D)
+    weights = np.ones_like(log_signal)
+
+    for i in range(max_iter):
+        # Weighted linear regression
+        model = sm.WLS(log_signal, bvals, weights=weights)
```
etpeterson, line 65: Here you're rolling your own iteratively re-weighted least squares. How does it compare to what is already in statsmodel? https://www.statsmodels.org/stable/generated/statsmodels.robust.robust_linear_model.RLM.html#statsmodels.robust.robust_linear_model.RLM

etpeterson (changes requested):

src/original/TF_reference/segmented_IVIMfit.py
Bneijndorff: I am no expert on this topic but what I found was that WLS can deal with heteroscedasticity and that RLM is less sensitive to outliers. Since the recommendation paper specifically stated WLS this is the one I used.

src/original/TF_reference/segmented_IVIMfit.py
Bneijndorff: I did some browsing indeed and this was the most straightforward implementation of a WLLS I could find

src/original/TF_reference/segmented_IVIMfit.py
oliverchampion: Hey Ben,

As far as I understand (https://www.sciencedirect.com/science/article/pii/S1053811913005223) the normal WLS is just a weighted least squares with pre-defined weights. 
Then, you can choose these in different ways (options a and b right before eq 7). 

Now, what we are implementing is slightly more sophisticated, an iterative WLS (eq 7). One thing the paper suggests is to use A or B to initiate the WLS (you now use ones). 

I think this iterative WLS is similar to statsmodels.robust.robust_linear_model.RLM. (the first sentence on its help page: "Robust Linear Model Estimate a robust linear model via **iteratively reweighted least squares** given a robust criterion estimator."

Possibly using a defined tool from Python may be more robust. However, if we really want to keep to the suggestions from this paper: https://www.sciencedirect.com/science/article/pii/S1053811913005223 we may need to implement it ourselves, as I cannot see how the weights are initiated and updated in the statsmodels package. However, your current implementation is not initiating the weights correctly either, so something may need to change :)

src/original/TF_reference/segmented_IVIMfit.py
oliverchampion: See earlier comment :)

src/original/TF_reference/segmented_IVIMfit.py
etpeterson: If the iterative route is the way, here's the documentation for the weighting functions you can choose in RLM: https://www.statsmodels.org/dev/rlm.html#norms

I think the re-weighting is also there, in the `scale_est` parameter. I think the `mad` option there might be somewhat similar to what you're using?

src/original/TF_reference/segmented_IVIMfit.py
```diff
@@ -0,0 +1,78 @@
...
+
+    bvalues_D: all bvalues above the threshold for fitting D
+
+    max_iter: the maximum number of iterations for WLS
+
+    tolerance: the tolerance for early stopping of the WLS
+    """
+
+    bvals = sm.add_constant(bvalues_D)
+    weights = np.ones_like(log_signal)
```
Bneijndorff, line 61: a

Bneijndorff:
I added weight initialization and updating based on (https://www.sciencedirect.com/science/article/pii/S1053811913005223) equation (7). The initial weights are based on the results of a LLS estimation, then the iterative WLLS starts and for iteration n the weights are based on the results of iteration n-1.

I did use my own implementation of the iterative WLLS since then we can precisely control how the weights are updated in each iteration.

Let me know what you think or if there should be some updates or changes.

I also see that unit-tests fail now for python versions 3.11 and higher do to some module import in scipy. I do not understand why this happens now since I did not change anything in the modules that are imported. If someone else has any idea please let me know!

Bneijndorff:
Would also be fine for me if someone else works on the RLM implementation, might indeed be interesting for the comparison.
Let's see if someone else has the time to do this :)

etpeterson:
So we're deferring the RLM change and doing that in a separate branch/PR?

oliverchampion:
if you agree :)

etpeterson:
Yeah, let's do it that way. Here's the link to the related feature request for later reference. https://github.com/OSIPI/TF2.4_IVIM-MRI_CodeCollection/issues/110

etpeterson (approved):

PR #104, opened 2025-05-28
cached the algorithms in pytest

oliverchampion:

For speed, I've cached the Pytest algorithms such that they are not re-initiated every run. Testing is faster now. Especially good for Matlab-based testing where the Matlab engine needs initiating (different PR)

none

### Checklist

- [ X ] Self-review of changed code
- [ X ] Added automated tests where applicable
- [ ] Update Docs & Guides

### DIFF
--- a/tests/IVIMmodels/unit_tests/test_ivim_fit.py
+++ b/tests/IVIMmodels/unit_tests/test_ivim_fit.py
@@ -29,6 +29,7 @@ def tolerances_helper(tolerances, data):
         tolerances["atol"] = tolerances.get("atol", {"f": 2e-1, "D": 5e-4, "Dp": 4e-2})
     return tolerances
 
+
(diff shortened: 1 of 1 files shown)


oliverchampion:
@etpeterson, with regards to (1) I will look into this. 
With regards to (2), this doesn't make a huge difference with just python code. But Matlab takes 20 seconds to initiate. If this is done for every single test point, this takes ages. Note that I only store algorithms with identical settings.
I would imagine that one would like the algorithm to remain in memory. If it does, we can fix that once it occurs?

oliverchampion:
It is the initiation of Matlab within python. I think it starts some sort
of Matlab kernel. I now intimate the python-matlab connection whenever the
algorithm is initiated (before it was on a per-fit basis).

etpeterson:
Does your other ticket make this one obsolete?

PR #106, opened 2025-06-13
Wrapper dev - attribute harmonization and signal normalization

IvanARashid:

Two things in this wrapper update:

1. Harmonization of standardized algorithm attributes. Some names have been changed, some have been added where missing.
2. Signal normalization before the fit. Has been added to osipi_fit() and osipi_fit_full_volume()

### Checklist

- [ ] Update Docs & Guides

### DIFF
--- a/src/wrappers/OsipiBase.py
+++ b/src/wrappers/OsipiBase.py
@@ -103,12 +103,14 @@ class OsipiBase:
             #args = [data[ijk], use_bvalues]
             #fit = list(self.ivim_fit(*args, **kwargs))
             #results[ijk] = fit
+        minimum_bvalue = np.min(use_bvalues) # We normalize the signal to the minimum bvalue. Should be 0 or very close to 0.
+        b0_indices = np.where(use_bvalues == minimum_bvalue)[0]
 
         for ijk in tqdm(np.ndindex(data.shape[:-1]), total=np.prod(data.shape[:-1])):
             # Normalize array
             single_voxel_data = data[ijk]
-            single_voxel_data_s0 = single_voxel_data[0]
-            single_voxel_data_normalized = single_voxel_data/single_voxel_data_s0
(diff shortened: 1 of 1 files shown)


src/wrappers/OsipiBase.py
```diff
@@ -105,7 +105,12 @@ def osipi_fit(self, data, bvalues=None, **kwargs):
             #results[ijk] = fit
 
         for ijk in tqdm(np.ndindex(data.shape[:-1]), total=np.prod(data.shape[:-1])):
-            args = [data[ijk], use_bvalues]
+            # Normalize array
+            single_voxel_data = data[ijk]
+            single_voxel_data_s0 = single_voxel_data[0]
```
etpeterson, line 110: This is assuming that the first index will be the b=0 data. Probably valid but is that an requirement we want to make? Seems like something someone is going to get caught on at some point.
I actually have an ancient PR that should find the b=0 automatically here: https://github.com/OSIPI/TF2.4_IVIM-MRI_CodeCollection/pull/80/files#diff-3e73b3e25749003f8536e5b52b164fa715d358ea4237220c3a1e32f445be03a7
Would dusting that off be good to include here?

etpeterson (changes requested):
Just the one comment really. Also wondering if this is something to get in before the paper?

src/wrappers/OsipiBase.py
IvanARashid: I agree with having something more sophisticated than this. I remember raising this and it sort of barely landed on us just requiring it to be first. But I like the solution of taking the mean of all b0's in this example, and shouldn't be hard to implement as we have all the b-values in an array. So I'll adjust the code to include this functionality.

oliverchampion:
Looks good to me! Happy to merge

etpeterson (approved):
Looks good

PR #116, opened 2025-09-15
docs: add note on DAG Installer Chart location and usage

Unique-Usman:
Added documentation pointing to the DAG Installer Chart in the Kaapana repository and clarified that users must download it locally and ensure the `requirements.yaml` path points to the correct directory before running workflows.


Unique-Usman:
@etpeterson - Small important update here.

PR #136, opened 2026-01-24
Feat: Implement Weighted Least Squares (WLS) algorithm

### LINKED ISSUE #110: [Feature] Weighted Least Squares algorithm
### Feature description

We are writing a recommendations paper recommending WLS fitting. Currently, Ben implemented a WLS based on the recommendations paper, but did this from scratch (no dedicated tools). I would be curious as to how "reproducible" our recommendations are (i.e. people who read the paper and implement the algorithm may still end up with different solutions). This would be very valuable information and would strongly point to the need for an open source repository (us!)

If you want to work on this, pop me (oliverchampion) a message and I can send you the wording in the recommendations paper.
 

### Describe the solution

A WLS in accoordance with the recommendations paper.

We could add a WLS recommendations approach based on this paper https://www.sciencedirect.com/science/article/pii/S1053811913005223 but using the build in WLS from statsmodel:

https://www.statsmodels.org/stable/generated/statsmodels.robust.robust_linear_model.RLM.html#statsmodels.robust.robust_linear_model.RLM

Comment from Eric:
If the iterative route is the way, here's the documentation for the weighting functions you can choose in RLM: https://www.statsmodels.org/dev/rlm.html#norms

I think the re-weighting is also there, in the scale_est parameter. I think the mad option there might be somewhat similar to what you're using?

### Describe alternatives

Anyone with alternative ways to implement the approach, please do! The more the merrier for our point.
E.g. in Matlab?

### Additional context

_No response_

### Are you working on this?

None

Aarshpatel12:
### Description
This PR implements the **Weighted Least Squares (WLS)** fitting algorithm using the `statsmodels` library, as requested in Issue #110.

### Key Changes
- **New Wrapper:** Added `src/standardized/OSIPI_Statsmodels_WLS.py`.
- **Algorithm:** Implemented a segmented fitting approach:
  1. Fits *D* and *f* using RLM on high b-values (>200 s/mm²).
  2. Fits *D** using RLM on the residuals (low b-values).
- **Unit Test:** Added `tests/IVIMmodels/unit_tests/test_OSIPI_Statsmodels_WLS.py` which verifies the fitting accuracy against synthetic ground-truth data.

### Dependencies
- Requires `statsmodels` (standard pip package).
- Compatible with `numpy < 2.0.0` (tested with 1.26.4).

### Related Issue
Fixes #110

### DIFF
--- a/src/standardized/OSIPI_Statsmodels_WLS.py
+++ b/src/standardized/OSIPI_Statsmodels_WLS.py
@@ -1,0 +1,138 @@
+from src.wrappers.OsipiBase import OsipiBase
+import numpy as np
+import importlib.util
+
+class OSIPI_Statsmodels_WLS(OsipiBase):
+    """
+    Wrapper for Weighted Least Squares (WLS) fitting using Statsmodels.
+    Based on Issue #110.
+    """
+    
+    # IVIM data is usually 4D (x, y, z, b-values) or 2D (voxels, b-values)
+    ACCEPTED_DIMENSIONS = (2, 4)
+    
+    def __init__(self, thresholds=None, bounds=None, initial_guess=None):
+        """
+        Initialize the wrapper and check for dependencies.
+        """
+        super(OSIPI_Statsmodels_WLS, self).__init__(
+            thresholds=thresholds,
+            bounds=bounds,
+            initial_guess=initial_guess
+        )
+        
+        # Check if statsmodels is installed
+        if importlib.util.find_spec("statsmodels") is None:
+            print("Warning: 'statsmodels' package not found. Please run: pip install statsmodels")
+
+    def osipi_fit(self, data, bvalues):
+        """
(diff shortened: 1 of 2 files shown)


oliverchampion:
Thanks for looking into this, and taking the effort to implement it in the standardized pipeline and think about testing!
However, the coding (predominantly the placing of code) is not in line with the repository. Could you please:
- Read the "how to commit" section on the read me
- Contributions must be place in a folder named "Initials_Institution" under src/original, e.g. "src/original/OGC_AmsterdamUMC" --> so this is where the fit code should be
- The fit code then can be called from main/src/standardized, but for this please copy an existing main/src/standardized implementation and adapt to your code, such that all variables are defined properly.
- Testing has already been fully automated, and will run by adding the fit name to "main/tests/IVIMmodels/unit_tests/algorithms.json". So you can add the name there and remove custom testing.

PR #147, opened 2026-02-27
fix: add missing D_and_Ds_swap() to 13 algorithms for consistent parameter labelling

### LINKED ISSUE #146: [BUG] Missing D_and_Ds_swap() post-fit correction in 13 out of 22 Python algorithms

### Describe the bug

The `D_and_Ds_swap()` method in `OsipiBase` corrects mislabelled D and D* (Dp) parameters after fitting. When the optimizer assigns them to the wrong parameter (i.e., `D > Dp` and `Dp < 0.05`), it swaps them and adjusts `f = 1 - f`.

Currently, only **9 out of 22** non-DL/non-MATLAB algorithms call this method after fitting. The remaining **13 algorithms** skip this correction, leading to **inconsistent parameter labelling** across algorithms when the optimizer produces swapped results.

### Screenshots [optional]

_No response_

### Steps To Reproduce

```python
import numpy as np
import os, sys
sys.path.insert(0, os.getcwd())

# Check source code for the presence of D_and_Ds_swap
algo_dir = os.path.join("src", "standardized")

algorithms = [
    "IAR_LU_biexp", "IAR_LU_segmented_2step", "IAR_LU_segmented_3step",
    "IAR_LU_subtracted", "IAR_LU_modified_mix", "IAR_LU_modified_topopro",
    "PV_MUMC_biexp",
    "OGC_AmsterdamUMC_biexp", "OGC_AmsterdamUMC_biexp_segmented", "OGC_AmsterdamUMC_Bayesian_biexp",
    "OJ_GU_seg", "PvH_KB_NKI_IVIMfit", "TF_reference_IVIMfit",
    "TCML_TechnionIIT_lsqlm", "TCML_TechnionIIT_lsqtrf", "TCML_TechnionIIT_lsqBOBYQA",
    "TCML_TechnionIIT_SLS", "TCML_TechnionIIT_lsq_sls_lm",
    "TCML_TechnionIIT_lsq_sls_trf", "TCML_TechnionIIT_lsq_sls_BOBYQA",
    "ETP_SRI_LinearFitting", "ASD_MemorialSloanKettering_QAMPER_IVIM",
]

for algo in sorted(algorithms):
    fpath = os.path.join(algo_dir, f"{algo}.py")
    if os.path.exists(fpath):
        with open(fpath) as f:
            content = f.read()
        has_it = "D_and_Ds_swap" in content
        tag = "HAS" if has_it else "MISSING"
        print(f"  [{tag:7s}] {algo}")
```

### Expected behavior

All algorithms should consistently apply `D_and_Ds_swap()` after fitting to ensure correct D/D* labelling.

### Additional context

**Output on `main` branch:**
```
  [HAS    ] ASD_MemorialSloanKettering_QAMPER_IVIM
  [HAS    ] ETP_SRI_LinearFitting
  [HAS    ] IAR_LU_biexp
  [HAS    ] IAR_LU_modified_mix
  [HAS    ] IAR_LU_modified_topopro
  [HAS    ] IAR_LU_segmented_2step
  [MISSING] IAR_LU_segmented_3step
  [MISSING] IAR_LU_subtracted
  [MISSING] OGC_AmsterdamUMC_Bayesian_biexp
  [MISSING] OGC_AmsterdamUMC_biexp
  [MISSING] OGC_AmsterdamUMC_biexp_segmented
  [MISSING] OJ_GU_seg
  [MISSING] PV_MUMC_biexp
  [MISSING] PvH_KB_NKI_IVIMfit
  [HAS    ] TCML_TechnionIIT_lsqBOBYQA
  [HAS    ] TCML_TechnionIIT_lsqlm
  [MISSING] TCML_TechnionIIT_lsq_sls_BOBYQA
  [MISSING] TCML_TechnionIIT_lsq_sls_lm
  [MISSING] TCML_TechnionIIT_lsq_sls_trf
  [HAS    ] TCML_TechnionIIT_lsqtrf
  [MISSING] TCML_TechnionIIT_SLS
  [MISSING] TF_reference_IVIMfit
```

13 algorithms are MISSING the call.

### Suggested Fix

Add `results = self.D_and_Ds_swap(results)` just before `return results` in the `ivim_fit()` method of each of the 13 missing algorithms.

### Are you working on this?

Yes

Devguru-codes:
Add self.D_and_Ds_swap(results) call before return in ivim_fit() for:
- IAR_LU_segmented_3step, IAR_LU_subtracted
- PV_MUMC_biexp
- OGC_AmsterdamUMC_biexp, OGC_AmsterdamUMC_biexp_segmented, OGC_AmsterdamUMC_Bayesian_biexp
- OJ_GU_seg, PvH_KB_NKI_IVIMfit, TF_reference_IVIMfit
- TCML_TechnionIIT_SLS, TCML_TechnionIIT_lsq_sls_lm/trf/BOBYQA

This corrects mislabelled D and D* when the optimizer swaps them. Previously only 9/22 algorithms had this correction.

No regressions: 1127 passed, 167 skipped, 22 xfailed, 6 xpassed.

Fixes #146

### Checklist

- [ ] Update Docs & Guides

### DIFF
--- a/src/standardized/IAR_LU_segmented_3step.py
+++ b/src/standardized/IAR_LU_segmented_3step.py
@@ -114,5 +114,6 @@ class IAR_LU_segmented_3step(OsipiBase):
         results["f"] = fit_results.model_params[1]
         results["Dp"] = fit_results.model_params[2]
         results["D"] = fit_results.model_params[3]
+        results = self.D_and_Ds_swap(results)
         
         return results
--- a/src/standardized/IAR_LU_subtracted.py
+++ b/src/standardized/IAR_LU_subtracted.py
@@ -112,5 +112,6 @@ class IAR_LU_subtracted(OsipiBase):
         results["f"] = fit_results.model_params[1]
         results["Dp"] = fit_results.model_params[2]
         results["D"] = fit_results.model_params[3]
+        results = self.D_and_Ds_swap(results)
         
         return results
--- a/src/standardized/OGC_AmsterdamUMC_Bayesian_biexp.py
+++ b/src/standardized/OGC_AmsterdamUMC_Bayesian_biexp.py
@@ -101,6 +101,7 @@ class OGC_AmsterdamUMC_Bayesian_biexp(OsipiBase):
         results["D"] = fit_results[0]
         results["f"] = fit_results[1]
         results["Dp"] = fit_results[2]
+        results = self.D_and_Ds_swap(results)
 
         return results
 
--- a/src/standardized/OGC_AmsterdamUMC_biexp.py
+++ b/src/standardized/OGC_AmsterdamUMC_biexp.py
@@ -73,5 +73,6 @@ class OGC_AmsterdamUMC_biexp(OsipiBase):
         results["D"] = fit_results[0]
         results["f"] = fit_results[1]
         results["Dp"] = fit_results[2]
+        results = self.D_and_Ds_swap(results)
 
         return results
--- a/src/standardized/OGC_AmsterdamUMC_biexp_segmented.py
+++ b/src/standardized/OGC_AmsterdamUMC_biexp_segmented.py
@@ -79,5 +79,6 @@ class OGC_AmsterdamUMC_biexp_segmented(OsipiBase):
         results["D"] = fit_results[0]
         results["f"] = fit_results[1]
         results["Dp"] = fit_results[2]
+        results = self.D_and_Ds_swap(results)
 
         return results
--- a/src/standardized/OJ_GU_seg.py
+++ b/src/standardized/OJ_GU_seg.py
@@ -76,5 +76,6 @@ class OJ_GU_seg(OsipiBase):
         results["f"] = fit_results['f']
         results["Dp"] = fit_results['Dstar']
         results["D"] = fit_results['D']
+        results = self.D_and_Ds_swap(results)
         
         return results
--- a/src/standardized/PV_MUMC_biexp.py
+++ b/src/standardized/PV_MUMC_biexp.py
@@ -77,5 +77,6 @@ class PV_MUMC_biexp(OsipiBase):
         results["f"] = fit_results[1]
         results["Dp"] = fit_results[2]
         results["D"] = fit_results[0]
+        results = self.D_and_Ds_swap(results)
         
         return results
--- a/src/standardized/PvH_KB_NKI_IVIMfit.py
+++ b/src/standardized/PvH_KB_NKI_IVIMfit.py
@@ -73,5 +73,6 @@ class PvH_KB_NKI_IVIMfit(OsipiBase):
         results["D"] = fit_results[0][0,0,0]/1000
         results["f"] = fit_results[1][0,0,0]
         results["Dp"] = fit_results[2][0,0,0]/1000
+        results = self.D_and_Ds_swap(results)
 
         return results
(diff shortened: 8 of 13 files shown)


Devguru-codes:
This PR adds the missing `D_and_Ds_swap()` call to 13 standardized algorithm implementations that were missing it. This method corrects mislabelled D and D* (Dp) parameters when the optimizer assigns them to the wrong parameter.

Previously, only 9 out of 22 Python algorithms applied this correction, leading to inconsistent parameter labelling.

### Changes Made

Added `results = self.D_and_Ds_swap(results)` before `return results` in the `ivim_fit()` method of these 13 files:

| # | File |
|---|---|
| 1 | `IAR_LU_segmented_3step.py` |
| 2 | `IAR_LU_subtracted.py` |
| 3 | `PV_MUMC_biexp.py` |
| 4 | `OGC_AmsterdamUMC_biexp.py` |
| 5 | `OGC_AmsterdamUMC_biexp_segmented.py` |
| 6 | `OGC_AmsterdamUMC_Bayesian_biexp.py` |
| 7 | `OJ_GU_seg.py` |
| 8 | `PvH_KB_NKI_IVIMfit.py` |
| 9 | `TF_reference_IVIMfit.py` |
| 10 | `TCML_TechnionIIT_SLS.py` |
| 11 | `TCML_TechnionIIT_lsq_sls_lm.py` |
| 12 | `TCML_TechnionIIT_lsq_sls_trf.py` |
| 13 | `TCML_TechnionIIT_lsq_sls_BOBYQA.py` |

Each change is a single +1 line addition. No other code was modified.

### Testing/Verification

- Ran the full test suite: **1127 passed, 167 skipped, 22 xfailed, 6 xpassed** — identical to main baseline
- **Zero regressions**. The `D_and_Ds_swap()` is a no-op when parameters are already correctly labelled, so existing test results are unchanged.

oliverchampion:
Some discussion and history here: 
If I remember correctly, in the past, I implemented the swap in the wrapper to make sure it hit all algorithms. However, this gave some errors for edge-cases (i.e. when f=0 or f=1, we get a mono-exponential signal, for which either D or D* become meaningless. This gave some weird results (ie f=0, D fits normally, D* becomes very very small --> D and D* get swapped, messing up everything). Therefore, we chose to only implement this for some algorithms that benefited. 

@IvanARashid do I remember this correctly?

Devguru-codes:
I didn't know about this. I think you are right about the edge case when f ≈ 0 (mono-exponential, no perfusion), D fits normally but D* becomes meaninglessly small (e.g. D* → 0.001). This triggers the swap condition (D > Dp AND Dp < 0.05), which incorrectly swaps D and D* and flips f, messing up an otherwise valid fit. I have 2 options here (I think and correct me if I am wrong) - 
First is revert this PR or I can add a guard the swap with an additional check to avoid the edge case like this - 
```python 
if results['D'] > results['Dp'] and results['Dp'] < 0.05 and results['f'] > 0.01:
    # only swap when there's a meaningful perfusion fraction
```
I will need your confirmation here like what to do.

IvanARashid:
Hi! So I think there has been a misunderstanding here.

It is indeed true that theoretically, D* should be greater than D. However, some of these algorithms work in specific steps that ensure that D and D* are fit using appropriate data. E.g. D* is fit using low b-values, D using high b-values. In these cases, when we get a D* < D, it is because of bad data, and not because the optimizer has swapped them. Therefore, it's a bit dangerous to use the swap function which only looks at the magnitude of these quantities and does not consider how they were fit.

Right now, the results are implemented in a way which considers how they were fit based on studying the original algorithm, which is the desired functionality.

However, this change is applicable to the algorithms that perform a bi-exponential fit in a single step, where both D and D* are fit simultaneously. If you can identify these, we could consider applying this change for those algorithms. I can for example say that the two IAR_LU algorithms are not in this category and should have the change reverted.

oliverchampion:
After internal discussion, we think it makes sense to only add the swap to algorithms that requiere it and leave this up to the discretion of the coders for those algorithms.

PR #148, opened 2026-03-01
feature: Add Weighted Least Squares (WLS) IVIM fitting algorithm — DT_IIITN (Feature #110)

### LINKED ISSUE #110: [Feature] Weighted Least Squares algorithm

Devguru-codes:
Hello @oliverchampion. The man who was assigned this issue didnt update his PR and had issues in it so I worked according to "how to commit" given in readme. I have also taken 1 decision explained in this PR ( `statsmodels.robust.robust_linear_model.RLM` vs numpy WSL) . If you suggest that we should use the former than latter - please let me know and I will update this PR.

### What this PR does

Closes #110. Implements a **Weighted Least Squares (WLS) segmented IVIM fitting algorithm** in accordance with the repository's contribution structure, following the approach recommended in Veraart et al. (2013).

This PR addresses the rejected PR #136, which implemented the same feature but placed code in the wrong location. This PR follows the correct structure:
- Fit code → `src/original/Initials_Institution/`
- Standardized wrapper → `src/standardized/`
- Test registration → `algorithms.json` (no custom test file needed)

---

### Algorithm

**Segmented two-step WLS approach:**

**Step 1 — Estimate D (high b-values, b ≥ 200 s/mm²):**
At high b-values the perfusion component decays to ~0, leaving:
```
ln(S(b)) ≈ ln(1 - f) - b·D
```
Weighted linear regression on log-signal with weights w = S(b)² (Veraart correction for heteroscedasticity introduced by the log-transform). Intercept yields f.

**Step 2 — Estimate D\* (low b-values):**
Subtract the diffusion component to isolate the perfusion signal:
```
residual(b) = S(b) - (1-f)·exp(-b·D)
ln(residual(b)) ≈ ln(f) - b·D*
```
Weighted linear regression with weights w = residual² gives D\*.

**Implementation:** Pure `numpy` (no extra dependencies). Uses the closed-form weighted normal equations:
```
β = (X^T W X)^{-1} X^T W y
```

---

### Files Changed

| Action | File | Description |
|---|---|---|
| **NEW** | `src/original/DT_IIITN/__init__.py` | Package marker |
| **NEW** | `src/original/DT_IIITN/wls_ivim_fitting.py` | Raw WLS algorithm (numpy only) |
| **NEW** | `src/standardized/DT_IIITN_WLS.py` | OsipiBase standardized wrapper |
| **MODIFY** | `tests/IVIMmodels/unit_tests/algorithms.json` | Registered `DT_IIITN_WLS` |

---

### API Usage

```python
import numpy as np

bvals = np.array([0, 10, 20, 50, 100, 200, 400, 800])

# Basic usage
fit = OsipiBase(algorithm="DT_IIITN_WLS", bvalues=bvals)
results = fit.ivim_fit(signal, bvals)
# → {"D": 0.0010, "f": 0.08, "Dp": 0.013}

# Custom segmentation threshold (default: 200 s/mm²)
fit = OsipiBase(algorithm="DT_IIITN_WLS", bvalues=bvals, thresholds=[150])

# Direct class import
from src.standardized.DT_IIITN_WLS import DT_IIITN_WLS
fit = DT_IIITN_WLS(bvalues=bvals)
```

---

### Design Decision: numpy WLS vs statsmodels RLM

The issue description suggests using `statsmodels.robust.robust_linear_model.RLM`. After evaluating both approaches:

| | statsmodels RLM | This PR (numpy WLS) |
|---|---|---|
| **Weighting** | Iterative reweighting (IRLS) via M-estimators | Closed-form S² weighting (Veraart 2013) |
| **Test suite runtime** | ~4+ hours (per-voxel IRLS is very slow) | ~17 minutes |
| **Extra dependency** | `statsmodels` | None (numpy only) |
| **Statistical basis** | Robust to outliers via IRLS | Optimal WLS for Rician-noise log-signal |

The numpy WLS approach implements the **same weighting strategy** recommended by Veraart et al. (w = S²) but using the closed-form normal equations — which is exactly what the paper validates. The `statsmodels.RLM` iterative approach adds robustness to outliers but at a ~15× runtime cost that is impractical for voxel-wise fitting on large datasets.

If you (anyone who will see my pr) prefers the full RLM approach, it can be easily switched in `wls_ivim_fitting.py` by replacing the `_weighted_linreg` call with `sm.RLM(...).fit()`.

---

### Reference

Veraart, J. et al. (2013). "Weighted linear least squares estimation of diffusion MRI parameters: strengths, limitations, and pitfalls." *NeuroImage*, 81, 335–346.
DOI: [10.1016/j.neuroimage.2013.05.028](https://doi.org/10.1016/j.neuroimage.2013.05.028)

---

### Testing

Testing is handled automatically by the existing framework via `algorithms.json` registration.

| Metric | Result |
|---|---|
| Total passed | **1184** (57 new tests for `DT_IIITN_WLS`) |
| Regressions | **0** |
| `test_ivim_fit_saved` | 593 passed, exit code 0 ✅ |
| `test_volume` / `test_parallel` errors | Pre-existing on `main` (Windows `FileNotFoundError` in subprocess, unrelated to this PR — confirmed by running same tests on `main` before this branch) |

Sample fit result on synthetic signal (`f=0.10, D=0.0010, D*=0.010`):
```
Fitted: f=0.080, D=0.0010, Dp=0.013  ✅ (expected for linearized WLS)
```

---

### Checklist

- [x] Follows `src/original/Initials_Institution/` + `src/standardized/` contribution structure
- [x] No custom test file — registered in `algorithms.json`
- [x] No new dependencies (`numpy` only — `statsmodels` evaluated but removed for performance; see Design Decision above)
- [x] Tested against automated test suite — zero regressions

### DIFF
--- a/src/standardized/DT_IIITN_WLS.py
+++ b/src/standardized/DT_IIITN_WLS.py
@@ -1,0 +1,87 @@
+from src.wrappers.OsipiBase import OsipiBase
+from src.original.DT_IIITN.wls_ivim_fitting import wls_ivim_fit
+import numpy as np
+
+
+class DT_IIITN_WLS(OsipiBase):
+    """
+    Weighted Least Squares IVIM fitting using statsmodels Robust Linear Model.
+
+    Segmented approach:
+    1. Estimate D from high b-values using robust linear regression on log-signal
+    2. Estimate D* from residuals at low b-values using robust linear regression
+
+    Author: Devguru Tiwari, IIIT Nagpur
+
+    Reference:
+        Veraart, J. et al. (2013). "Weighted linear least squares estimation of
+        diffusion MRI parameters: strengths, limitations, and pitfalls."
+        NeuroImage, 81, 335-346.
+        DOI: 10.1016/j.neuroimage.2013.05.028
+    """
+
+    # Algorithm identification
+    id_author = "Devguru Tiwari, IIIT Nagpur"
+    id_algorithm_type = "Weighted least squares segmented fit"
+    id_return_parameters = "f, D*, D"
+    id_units = "seconds per milli metre squared or milliseconds per micro metre squared"
+    id_ref = "https://doi.org/10.1016/j.neuroimage.2013.05.028"
+
+    # Algorithm requirements
+    required_bvalues = 4
+    required_thresholds = [0, 0]
+    required_bounds = False
+    required_bounds_optional = True
+    required_initial_guess = False
+    required_initial_guess_optional = True
+
+    # Supported inputs
+    supported_bounds = False
+    supported_initial_guess = False
+    supported_thresholds = True
+    supported_dimensions = 1
+    supported_priors = False
+
+    def __init__(self, bvalues=None, thresholds=None,
+                 bounds=None, initial_guess=None):
+        """
+        Initialize the WLS IVIM fitting algorithm.
+
+        Args:
+            bvalues (array-like, optional): b-values for the fitted signals.
+            thresholds (array-like, optional): Threshold b-value for segmented
+                fitting. The first value is used as the cutoff between high
+                and low b-values. Default: 200 s/mm².
+            bounds (dict, optional): Not used by this algorithm.
+            initial_guess (dict, optional): Not used by this algorithm.
+        """
+        super(DT_IIITN_WLS, self).__init__(
+            bvalues=bvalues, bounds=bounds,
+            initial_guess=initial_guess, thresholds=thresholds
+        )
+
+    def ivim_fit(self, signals, bvalues, **kwargs):
+        """Perform the IVIM fit using WLS.
+
+        Args:
+            signals (array-like): Signal intensities at each b-value.
+            bvalues (array-like, optional): b-values for the signals.
+
+        Returns:
+            dict: Dictionary with keys "D", "f", "Dp".
+        """
+        bvalues = np.array(bvalues)
+
+        # Use threshold as cutoff if available
+        cutoff = 200
+        if self.thresholds is not None and len(self.thresholds) > 0:
+            cutoff = self.thresholds[0]
+
+        D, f, Dp = wls_ivim_fit(bvalues, signals, cutoff=cutoff)
+
+        results = {}
+        results["D"] = D
+        results["f"] = f
+        results["Dp"] = Dp
+
+        return results
--- a/src/standardized/DT_IIITN_WLS.py
+++ b/src/standardized/DT_IIITN_WLS.py
@@ -5,11 +5,15 @@ import numpy as np
 
 class DT_IIITN_WLS(OsipiBase):
     """
-    Weighted Least Squares IVIM fitting using statsmodels Robust Linear Model.
+    Segmented IVIM fitting with selectable regression method.
+
+    Two methods are available:
+    - WLS: Weighted Least Squares with Veraart S² weights (default)
+    - RLM: Robust Linear Model with Huber's T norm (statsmodels)
 
     Segmented approach:
-    1. Estimate D from high b-values using robust linear regression on log-signal
-    2. Estimate D* from residuals at low b-values using robust linear regression
+    1. Estimate D from high b-values using linear regression on log-signal
+    2. Estimate D* from residuals at low b-values using linear regression
 
     Author: Devguru Tiwari, IIIT Nagpur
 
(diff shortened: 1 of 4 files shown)


src/standardized/DT_IIITN_WLS.py
```diff
@@ -0,0 +1,87 @@
...
+    id_algorithm_type = "Weighted least squares segmented fit"
+    id_return_parameters = "f, D*, D"
+    id_units = "seconds per milli metre squared or milliseconds per micro metre squared"
+    id_ref = "https://doi.org/10.1016/j.neuroimage.2013.05.028"
+
+    # Algorithm requirements
+    required_bvalues = 4
+    required_thresholds = [0, 0]
+    required_bounds = False
+    required_bounds_optional = True
```
oliverchampion: @IvanARashid , I'm a bit confused what the difference is between
requiered bounds
requiered bounds optional
supported bounds

Does it make sense for the second to be true and the others false?

Devguru-codes: I followed the src\standardized\ETP_SRI_LinearFitting.py file pattern which also has required_bounds_optional = True with supported_bounds = False. From what i understood, required_bounds_optional = True means means the wrapper's init file will accept bounds without raising an error, while supported_bounds = False means the underlying algorithm doesn't actually use them. If any suggestion come up, I will update this. Thank you.

etpeterson:
@Devguru-codes You say it's easy to swap between WLS and RLM. Would it be too hard to allow either one to be used by way of an input?

Devguru-codes:
> @Devguru-codes You say it's easy to swap between WLS and RLM. Would it be too hard to allow either one to be used by way of an input?

I have implemented it now by adding method parameter, the user can do method="RLM" as an input to seamlessly swap between the two. Thank you.

src/standardized/DT_IIITN_WLS.py
IvanARashid: Yeah, not the wisest naming convention on my part. required_bounds can be False while required_bounds_optional can be true for algorithms that e.g. don't require bounds but can still take them as inputs. But I think both should be false for ETP_SRI_LinearFitting since it doesn't look like any bounds are being passed to the algorithm. It's something that has been overlooked in our previous passes.

So required_bounds_optional just means if bounds are optional and can still be passed.

IvanARashid: Keep in mind that LinearFitting was one of the very first algorithms to be committed to the repo and standardized, so there can be variables there that are deprecated :)

Devguru-codes: Thank you @IvanARashid sir for helping here. My algorithm here doesnt use bounds and initial guesses, I will set required_bounds_optional = False and required_initial_guess_optional = False as well. Thank you.

Devguru-codes:
I have implemented all the changes suggested by experts. Please check it @oliverchampion sir. Thank you.

src/standardized/DT_IIITN_WLS.py
Devguru-codes: > Keep in mind that LinearFitting was one of the very first algorithms to be committed to the repo and standardized, so there can be variables there that are deprecated :)

Do we need to update it or should we let it be. If you think I should update it, please tell. I will do it in a new PR. Thank you.

oliverchampion:
Will merge after we publish the version of the Github accompanying our paper. Hopefully later today

PR #149, opened 2026-03-01
feat: Add body-part aware initial guesses for IVIM fitting (Feature #87)

Devguru-codes:
Hello @etpeterson - I saw no updates were made on this issue so I worked on it. If there is need of improvement, comment it and i will update this pr.

### Description

Closes #87. Adds organ-specific initial guesses, bounds, and thresholds for IVIM fitting, sourced from peer-reviewed literature. This enables faster convergence, reduced local minima, and more physiologically plausible results when scanning specific body parts.

### Problem

The current generic defaults (`f=0.1, D=0.001, Dp=0.01`) are reasonable for the brain but significantly off for other organs:

| Parameter | Current Generic | Brain (actual) | Liver (actual) | Kidney (actual) |
|---|---|---|---|---|
| **f** | 0.10 | 0.03–0.06 | 0.10–0.15 | 0.18–0.25 |
| **D** (×10⁻³) | 1.00 | 0.70–0.80 | 0.96–1.19 | 1.78–1.89 |
| **Dp** (×10⁻³) | 10.0 | 7–25 | **40–80** ❌ 6× off | **25–30** ❌ 3× off |

### Solution: Literature-Sourced Lookup Table

| Body Part | f | D (×10⁻³) | Dp (×10⁻³) | Source |
|---|---|---|---|---|
| **Brain** | 0.05 | 0.80 | 10.0 | Federau 2017 ([DOI](https://doi.org/10.1002/nbm.3780)) |
| **Liver** | 0.12 | 1.00 | 60.0 | Dyvorne 2013 ([DOI](https://doi.org/10.1016/j.ejrad.2013.03.003)), Guiu 2012 ([DOI](https://doi.org/10.1002/jmri.23762)) |
| **Kidney** | 0.20 | 1.90 | 30.0 | Li 2017 ([DOI](https://doi.org/10.1002/jmri.25571)), Ljimani 2020 ([DOI](https://doi.org/10.1007/s10334-019-00802-0)) |
| **Prostate** | 0.08 | 1.50 | 25.0 | Kuru 2014 ([DOI](https://doi.org/10.1007/s00330-014-3165-y)) |
| **Pancreas** | 0.18 | 1.20 | 20.0 | Barbieri 2020 ([DOI](https://doi.org/10.1002/mrm.27910)) |
| **Head & Neck** | 0.15 | 1.00 | 25.0 | Sumi 2012 ([DOI](https://doi.org/10.1259/dmfr/15696758)) |
[...]

PR #154, opened 2026-04-15
fix conftest boolean option

DKuppens:

Fixes the boolean option in the conftest. The old implementation was set to True if anything was supplied to the --ricianNoise or --usePrior (also if you supply the False argument). To fix this, the new implementation is set to False as default but when --ricianNoise or --usePrior flags are supplied is set to True.

### Checklist

- [ x] Self-review of changed code
- [ x] Added automated tests where applicable
- [ NA] Update Docs & Guides

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -19,14 +19,14 @@ def pytest_addoption(parser):
     )
     parser.addoption(
         "--ricianNoise",
+        action="store_true",
         default=False,
-        type=bool,
         help="Use Rician noise, non-rician is gaussian",
     )
     parser.addoption(
         "--usePrior",
+        action="store_true",
         default=False,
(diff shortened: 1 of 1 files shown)


oliverchampion:
Hey Daan, this seems to be failing some tests...

Devguru-codes:
> Hey Daan, this seems to be failing some tests...

I believe the macOS failures are not from the change, they are pre-existing CI issues:
1. 3.12: dipy needs libomp (brew install libomp) on macOS
2. 3.11/3.13: PyTorch CPU segfault in test_deep_learning_algorithms
I have fixed both in PR #149. Once that merges, rebasing your branch should make these pass. Thank you.

PR #160, opened 2026-05-04
Fix: Add algorithm name validation in OsipiBase.osipi_initiate_algorithm()

### LINKED ISSUE #159: [BUG] No error handling or validation in `OsipiBase.__init__()` for invalid algorithm names

### Describe the bug

When initializing `OsipiBase` with a misspelled or invalid algorithm name, the dynamic class loader fails with a confusing, low-level Python exception (`ModuleNotFoundError` or `AttributeError`) rather than providing a clear error message to the user.

Because `OsipiBase` dynamically imports the algorithm file based on the string name (via `importlib`), passing a typo (like `"IAR_LU_biepx"` instead of `"IAR_LU_biexp"`) leads directly to an `importlib` crash. There is no prior validation of the algorithm name against the known list of algorithms in `algorithms.json` or `src/standardized/`.

### Screenshots [optional]

_No response_

### Steps To Reproduce

1. Create the following simple python script to initialize `OsipiBase` with a typo:
```python

# Note the typo: 'biepx' instead of 'biexp'
fit = OsipiBase(algorithm="IAR_LU_biepx")
```
2. Run the script.
3. Observe the confusing `ModuleNotFoundError`:
```
Traceback (most recent call last):
  ...
  File "...\src\wrappers\OsipiBase.py", line 160, in osipi_initiate_algorithm
    self.__class__ = getattr(importlib.import_module(import_path), algorithm)
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'src.standardized.IAR_LU_biepx'
```

### Expected behavior

`OsipiBase.__init__()` should validate the requested algorithm string before attempting the dynamic import. 

If the algorithm name is not recognized, it should raise a clear, user-friendly `ValueError` (e.g., `"ValueError: Algorithm 'IAR_LU_biepx' not found. Available algorithms are: [...]"`).

### Additional context

- This makes the library harder to use for newcomers who might make simple typos.
- Adding this validation would make the API much more robust and self-documenting.
- Partially related to Issue #45 (implement requirement checks).
- The `src/standardized/` directory already contains all valid algorithm `.py` files and can be used as the source of truth for validation.

### Are you working on this?

Yes

Devguru-codes:
This PR fixes the issue where `OsipiBase` crashes with a confusing `ModuleNotFoundError` or `AttributeError` when initialized with a misspelled or invalid algorithm name. The dynamic class loader in `osipi_initiate_algorithm()` previously performed **zero validation** before calling `importlib.import_module()`, causing raw Python import errors to surface to the user.

### `src/wrappers/OsipiBase.py`

1. **Algorithm name validation**: Before attempting the dynamic import, the method now scans the `src/standardized/` directory using `pathlib.glob("*.py")` to build a set of valid algorithm names at runtime. This approach is always in sync with the actual codebase — no hardcoded list needed.

2. **Clear `ValueError`**: If the requested algorithm is not found, a `ValueError` is raised listing all available algorithms alphabetically, making typos easy to spot.

3. **Secondary safety net**: The `importlib.import_module()` and `getattr()` calls are now wrapped in explicit `try/except` blocks with descriptive error messages:
   - `ImportError` — if the file exists but fails to import (e.g., syntax error in the module).
   - `AttributeError` — if the module loads but doesn't contain the expected class name.

4. **Updated docstring**: Added `Raises` section documenting the `ValueError`.

### `tests/IVIMmodels/unit_tests/test_ivim_fit.py`

Added `test_invalid_algorithm_name()` regression test with 3 sub-checks:

1. A typo (`"IAR_LU_biepx"`) raises `ValueError` with `"not found"` in the message.
2. A completely fake name (`"totally_fake_algorithm"`) raises `ValueError` that includes `"Available algorithms are:"` and lists known algorithms like `IAR_LU_biexp` and `PV_MUMC_biexp`.
3. A valid algorithm (`"IAR_LU_biexp"`) still initializes successfully without errors (no false positives).

## Testing & Validation

```
$ python -m pytest tests/IVIMmodels/unit_tests/test_ivim_fit.py::test_invalid_algorithm_name -v
tests/.../test_ivim_fit.py::test_invalid_algorithm_name PASSED [100%]
1 passed in 1.81s

$ python -m pytest tests/IVIMmodels/unit_tests/test_ivim_fit.py::test_default_bounds_and_initial_guesses -v
22 passed, 5 skipped in 2.24s
```

-  **New test passes** — invalid names correctly raise `ValueError`
-  **Existing tests unaffected** — all 22 algorithm-loading tests still pass (5 MATLAB/DL skipped as expected)

fixes #159

### DIFF
--- a/src/wrappers/OsipiBase.py
+++ b/src/wrappers/OsipiBase.py
@@ -154,8 +154,10 @@ class OsipiBase:
         # Import the algorithm
         root_path = pathlib.Path(__file__).resolve().parents[2]
         if str(root_path) not in sys.path:
-            print("Root folder not in PYTHONPATH")
-            return False
+            raise RuntimeError(
+                "Root folder not found in PYTHONPATH. "
+                "Please ensure the project root is in sys.path."
+            )
 
         # ------------------------------------------------------------------
         # Validate algorithm name against available modules in src/standardized/
--- a/src/wrappers/OsipiBase.py
+++ b/src/wrappers/OsipiBase.py
@@ -145,6 +145,10 @@ class OsipiBase:
 
         Args:
             algorithm (string): The name of the algorithm, should be the same as the file in the src/standardized folder without the .py extension.
+        
+        Raises:
+            ValueError: If the algorithm name does not correspond to any .py file
+                        in ``src/standardized/``.
         """
         
         # Import the algorithm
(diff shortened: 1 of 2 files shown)


src/wrappers/OsipiBase.py
```diff
@@ -145,19 +145,58 @@ def osipi_initiate_algorithm(self, algorithm, **kwargs):
...
+        Raises:
+            ValueError: If the algorithm name does not correspond to any .py file
+                        in ``src/standardized/``.
         """
         
         # Import the algorithm
         root_path = pathlib.Path(__file__).resolve().parents[2]
         if str(root_path) not in sys.path:
             print("Root folder not in PYTHONPATH")
             return False
```
IvanARashid: Could you also change this False into an exception here?

IvanARashid (changes requested):
Looks good, just one more adjustment so that we stay consistent.

src/wrappers/OsipiBase.py
Devguru-codes: Ok

Devguru-codes:
> Looks good, just one more adjustment so that we stay consistent.

I have implemented the change sir. Thank you.

oliverchampion:
Looks good to me. If Ivan agrees we can merge

IvanARashid (approved):
Looks good now, thanks!

PR #161, opened 2026-05-05
Preproc-main into main

IvanARashid:

Added fitting and preprocessing folders to /src/original, with corresponding fixes to the imports in various files in the repository.

Added a line to pytest.ini to ignore files starting with "wip_*" during automatic testing.


phantoms/brain/sim_brain_phantom_preproc.py
```diff
@@ -0,0 +1,82 @@
+import os
```
oliverchampion: Is it needed to have a seperate file for the brain phantom for preprocessing? Can this not be combined with Oscar's digital brain phantom with additional "default" options for turning preprocessing-specific things on/off? This would make the code cleaner and phantoms more alligned

oliverchampion (commented):
With regards to the Matlab codes --> do we need to include all toolboxes like nifti_toolbox? This could make submits unnesicarily long. Also should check what the licensing is there; are we allowed to share that code? are we allowed to share it with out own license (typically licensing only allows sharing code with the same license).

phantoms/brain/sim_brain_phantom_preproc.py
IvanARashid: The new file kind of looks better, it might make the existing one redundant. But I'm not sure. @ElinaPetersson ?

IvanARashid:
> With regards to the Matlab codes --> do we need to include all toolboxes like nifti_toolbox? This could make submits unnesicarily long. Also should check what the licensing is there; are we allowed to share that code? are we allowed to share it with out own license (typically licensing only allows sharing code with the same license).

The stuff that was submitted by ASD_MemorialSloanKettering? It's part of fitting code and has been in the repo for a while now

phantoms/brain/sim_brain_phantom_preproc.py
ElinaPetersson: Yes, sounds reasonable. I'll make one script with all functionality but optional relaxation decay.

oliverchampion:
Fine to me once merge conflicts are resolved!

PR #165, opened 2026-05-09
centralize b-value management at initialization (#155)

### LINKED ISSUE #155: [Feature] Changing b-values
### Feature description

So for many algorithms, we now allow to give b-values at initialisation and at fit-time. This is confusing and probably is not supported for all algorithms

### Describe the solution

I think it is cleaner to only allow to give b-values at initialisation, as some algorithms then initialise e.g. priors, and changing b-values may rise issues. On the other side, for some specific applications, this may require re-initialising the algorithm unnecessarily often.

Alternatively, we can support changing b-values at fit-time , but we would have to add unit tests to see whether this works well.

### Describe alternatives

_No response_

### Additional context

_No response_

### Are you working on this?

None

Devguru-codes:
- Remove bvalues parameter from all 27 algorithm ivim_fit() and ivim_fit_full_volume() signatures in src/standardized/
- All algorithms now access b-values via self.bvalues exclusively
- Add **kwargs to algorithms that lacked it (OJ_GU_seg, PV_MUMC, PvH_KB_NKI, TF_reference)
- OsipiBase.osipi_fit() and osipi_fit_full_volume() enforce ValueError if bvalues not set at init
- Update all test files to pass bvalues at OsipiBase(...) init
- Update WrapImage production wrappers to use new API
- Remove now-redundant bvalue equality checks in DL algorithms

fixes #155

### DIFF
--- a/src/standardized/ASD_MemorialSloanKettering_QAMPER_IVIM.py
+++ b/src/standardized/ASD_MemorialSloanKettering_QAMPER_IVIM.py
@@ -80,11 +80,10 @@ class ASD_MemorialSloanKettering_QAMPER_IVIM(OsipiBase):
 
         initial_guess = [self.initial_guess["D"], self.initial_guess["f"], self.initial_guess["Dp"], self.initial_guess["S0"]]
 
-        bvalues=self.bvalues
         LB = np.array(bounds[0])[[1,0,2,3]]
         UB = np.array(bounds[1])[[1,0,2,3]]
 
-        fit_results = self.algorithm(np.array(signals)[:,np.newaxis], bvalues, LB, UB, np.array(initial_guess)[[1,0,2,3]])
+        fit_results = self.algorithm(np.array(signals)[:,np.newaxis], self.bvalues, LB, UB, np.array(initial_guess)[[1,0,2,3]])
 
         results = {}
         results["D"] = fit_results[0]
--- a/src/standardized/ASD_MemorialSloanKettering_QAMPER_IVIM.py
+++ b/src/standardized/ASD_MemorialSloanKettering_QAMPER_IVIM.py
@@ -80,7 +80,7 @@ class ASD_MemorialSloanKettering_QAMPER_IVIM(OsipiBase):
 
         initial_guess = [self.initial_guess["D"], self.initial_guess["f"], self.initial_guess["Dp"], self.initial_guess["S0"]]
 
-        bvalues=np.array(self.bvalues)
+        bvalues=self.bvalues
         LB = np.array(bounds[0])[[1,0,2,3]]
         UB = np.array(bounds[1])[[1,0,2,3]]
 
--- a/src/standardized/ASD_MemorialSloanKettering_QAMPER_IVIM.py
+++ b/src/standardized/ASD_MemorialSloanKettering_QAMPER_IVIM.py
@@ -66,12 +66,11 @@ class ASD_MemorialSloanKettering_QAMPER_IVIM(OsipiBase):
         (f_arr, D_arr, Dx_arr, s0_arr, fitted_dwi_arr, RSS, rms_val, chi, AIC, BIC, R_sq) = results
         return D_arr/1000, f_arr, Dx_arr/1000, s0_arr
 
-    def ivim_fit(self, signals, bvalues, **kwargs):
+    def ivim_fit(self, signals, **kwargs):
         """Perform the IVIM fit
 
         Args:
             signals (array-like)
-            bvalues (array-like, optional): b-values for the signals. If None, self.bvalues will be used. Default is None.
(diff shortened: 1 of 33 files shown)


Devguru-codes:
Hello @oliverchampion sir, please review the changes. If anything needs to be updated, let me know. Thank you.

src/standardized/IAR_LU_biexp.py
```diff
@@ -125,16 +120,14 @@ def ivim_fit_full_volume(self, signals, bvalues, **kwargs):
                        [self.bounds["S0"][1], self.bounds["f"][1], self.bounds["Dp"][1], self.bounds["D"][1]]]
         initial_guess = [self.initial_guess["S0"], self.initial_guess["f"], self.initial_guess["Dp"], self.initial_guess["D"]]
         if self.IAR_algorithm is None:
-            if bvalues is None:
-                bvalues = self.bvalues
-            else:
-                bvalues = np.asarray(bvalues)
+            bvalues = np.asarray(self.bvalues)
```
oliverchampion: Given the if statement above, we probably do not need the np.asarray. self.bvalues has alrady been set in an arrray.

oliverchampion: (for all IAR's algorithms)

src/standardized/ASD_MemorialSloanKettering_QAMPER_IVIM.py
```diff
@@ -81,7 +80,7 @@ def ivim_fit(self, signals, bvalues, **kwargs):
 
         initial_guess = [self.initial_guess["D"], self.initial_guess["f"], self.initial_guess["Dp"], self.initial_guess["S0"]]
 
-        bvalues=np.array(bvalues)
+        bvalues=np.array(self.bvalues)
```
oliverchampion: For all algorithms: do we need to set it als bvalues=np.array(self.bvalues)? Or could we not just use self.bvalues instead of bvalues?

src/wrappers/OsipiBase.py
```diff
@@ -279,40 +275,18 @@ def osipi_fit(self, data, bvalues=None, njobs=1, **kwargs):
...
-        
-        #args = [data, use_bvalues, use_initial_guess, use_bounds, use_thresholds]
-        #args = [arg for arg in args if arg is not None]
+        # b-values must be provided at initialization, not at fit-time.
+        if self.bvalues is None:
+            raise ValueError(
+                "b-values must be provided at initialization (__init__), not at fit-time. "
+                "Pass bvalues to the constructor: OsipiBase(bvalues=..., algorithm=...)"
+            )
+        use_bvalues = np.asarray(self.bvalues)
```
oliverchampion: I think this means we can skip all other np.array/np.asarray

oliverchampion:
Otherwise it looks good to me. I'm a little worried that stuf down the line may break, such as the longitudinal testing and the website deployment; these are only run once merged. But let's cross that bridge when we get there.

Maybe @IvanARashid can check the osipi_base.

src/standardized/ASD_MemorialSloanKettering_QAMPER_IVIM.py
Devguru-codes: i have updated it to self.bvalues.

src/wrappers/OsipiBase.py
Devguru-codes: yes, i did that now.

Devguru-codes:
@oliverchampion sir, i have implemented the suggested changes. please review it. Thank you.

src/standardized/IAR_LU_subtracted.py
oliverchampion: For all code, instead of stating

bvalues=self.bvalues
remaining_code(bvalues)

is it not better to just do:

remaining_code(self.bvalues)

oliverchampion: (for all standardized functions)

src/standardized/ASD_MemorialSloanKettering_QAMPER_IVIM.py
oliverchampion: this comment was either misinterpretted, or ignored. See comment below. But basically, do we need to define bvalues, or can we use self.bvalues directly?

Devguru-codes: I misinterpreted it, my bad.

Devguru-codes:
@oliverchampion sir,I have just pushed an update removing all the bvalues = self.bvalues. Since self.bvalues isn't mutated anywhere, replacing the local variable and using self.bvalues directly everywhere worked perfectly. I have checked it on my end. Thank you.

PR #167, opened 2026-05-10
replace bare except: with specific exception types

### LINKED ISSUE #166: [BUG] Bare except: blocks in original algorithm implementations

### Describe the bug

All 14 bare except: blocks catch every exception including KeyboardInterrupt, SystemExit, MemoryError, etc. This:
1. Makes debugging impossible - errors are silently swallowed
2. Violates PEP 8 ("Never use a bare except:")
3. Hides real bugs (e.g. TypeError, ValueError from bad input)

### Screenshots [optional]

_No response_

### Steps To Reproduce

- src/original/fitting/OGC_AmsterdamUMC/LSQ_fitting.py -13 occurrences
- src/original/fitting/PV_MUMC/triexp_fitting_algorithms.py -1 occurrence

### Expected behavior

Help in debugging the errors

### Additional context

Replacing each bare except: with the most specific exception type(s) that the try block can raise.

### Are you working on this?

Yes

Devguru-codes:
Replace all 13 bare except: blocks in LSQ_fitting.py and 1 in triexp_fitting_algorithms.py with specific exception types:

- Parallel fallbacks: except Exception
- curve_fit/minimize failures: except (RuntimeError, ValueError)
- Bayesian fit: except (RuntimeError, ValueError, TypeError)
- Goodness-of-fit: except (IndexError, ValueError)
- NNLS: except Exception

Also fix pre-existing bug: raise(params.message) on line 668 was raising a string (TypeError), not an exception. Changed to raise RuntimeError(params.message).

Fixes #166

### DIFF
--- a/src/original/fitting/OGC_AmsterdamUMC/LSQ_fitting.py
+++ b/src/original/fitting/OGC_AmsterdamUMC/LSQ_fitting.py
@@ -94,7 +94,7 @@ def fit_segmented_array(bvalues, dw_data, njobs=4, bounds=([0, 0, 0.005],[0.005,
 
             output = Parallel(n_jobs=njobs)(delayed(parfun)(i) for i in tqdm(range(len(dw_data)), position=0, leave=True))
             Dt, Fp, Dp = np.transpose(output)
-        except:
+        except Exception:
             # if fails, retry using single core
             single = True
     else:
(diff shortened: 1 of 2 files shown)


oliverchampion:
It is a while back for me, but I these exceptions were there to make sure the whole program does not fall down if e.g. it encounters a NAN or inf, or some other unexpected insert in one of the voxels. But I agree that the program should break if there is something fundamentally wrong.

Are we sure these are the only exceptions we want to catch to ensure stable behaviour?

Devguru-codes:
Hello @oliverchampion sir, i apologise for the delay, i was busy finding an internship. I got the time now and reviewed what you said. You were right, i did miss some exceptions. To find everything i wrote a sanity-check script that feeds nan, inf, zeros, huge values, negative signals, and other edge cases into every function. Result was 53/53 bad-data scenarios handled gracefully (no crashes).
I also added some more to be just safe -
a. FloatingPointError on curve_fit blocks (covers NaN/inf when numpy FP error mode is strict)
b. TypeError on goodness_of_fit (len() on a scalar raises TypeError, not IndexError)
c. TypeError on fit_segmented_tri_exp (there's a pre-existing bug on line 454: min(0.1) always raises TypeError since min() needs an iterable. The old bare except: was hiding this.)
I also tested with np.seterr(all='raise') to simulate the strictest floating-point mode. Everything degrades gracefully. Thank you.

PR #173, opened 2026-08-24
minor adaptations for rebuttal

oliverchampion:

Expanded osipibase.cite and added note about licenses.

### Checklist

- [ x  ] Self-review of changed code
- [ x ] Added automated tests where applicable
- [ x ] Update Docs & Guides


src/wrappers/OsipiBase.py
```diff
@@ -584,9 +583,94 @@ def D_and_Ds_swap(self,results):
...
+            ig = self.initial_guess
+            if isinstance(ig, dict):
+                parts = [f"{k} = {v}" for k, v in ig.items()]
+            else:
+                parts = [str(v) for v in ig]
+            defaulted = getattr(self, "forced_default_initial_guess", False)
+            bp_note = (
+                f" for {self.body_part} tissue" if defaulted and getattr(self, "body_part", None) else ""
+            )
+            source_note = " (default values)" if defaulted else ""
```
IvanARashid, line 640: Should we clarify that it is our default values? E.g. "OSIPI default values" instead of just "default values"? Just to differentiate them from the defaults of the original submission.

src/wrappers/OsipiBase.py
```diff
@@ -584,9 +583,94 @@ def D_and_Ds_swap(self,results):
...
+            ", implemented within the OSIPI IVIM code repository "
+            "(An open-source code repository for intravoxel incoherent motion analysis: ISMRM Open Science Initiative for Perfusion Imaging (OSIPI); "
+            "Jalnefjord et al., MRM 2026, in press)."
+        )
+        lines.append(method_sentence)
+
+        if self.id_ref:
+            lines.append(
+                f"This implementation should additionally be cited according to its "
+                f"original publication: {self.id_ref}.\n"
```
IvanARashid, line 619: When I ran .cite() and got the link, the ".\n" was included as part of the doi, which made the link not work as it unintentionally includes additional characters into the doi.

E.g. try running algorithm2.cite() in the intro notebook to reproduce.

I'm not really sure how the link is implemented, perhaps VS Code just creates it itself. But even so, the \n is there in plain text and it looks like the period becomes part of the link, so we might need some reformulation of the sentence. Perhaps put the doi inside parentheses? E.g. "... be cited according to its original publication (self.id_ref)"

IvanARashid: Image

IvanARashid (changes requested):
A couple of small things.

I really like the methods report!

oliverchampion:
I think these issues are now resolved

oscarjalnefjord:
Looks fine to me.  You might want to add units för D and Dp as it is printed for the b-values.

PR #177, opened 2026-09-21
Add EPI distortion phantom NIfTI data

siria98:
Added a QIBA diffusion phantom acquired on a 3T scanner to the phantoms folder. Includes two NIfTI acquisitions with opposite phase-encoding directions intended for testing EPI distortion correction pipelines, along with b-values, gradient directions and a README.


IvanARashid:
@oliverchampion So these are niftis, which we should probably have on zenodo with the other data? Or should we look into creating a separate data repo for the preprocessing stuff?

Then the question arises, will we be able to do automated testing if we have to download data from zenodo? We sort of end up with the problem that any kind of preprocessing will need to be tested on some kind of image array, so perhaps automated testing on github is not something we can do, and that we will need to move to something that is run offline and spits out a report that we can review instead...

oliverchampion:
@siria98 in the past we added data directly to github and it meant within days we got emails from github stating we used too much traffic. I.e. everytime someone downloads github, they download the phantom. 

We now have the phantoms in Zenodo, and download them locally whenever we need them. So I think we need to get rid of the nifti you uploaded on github and put it in Zenodo. In the past, removing it was a little challenging as it sits in the git history taking up bandwidth. @DKuppens, do you know how we did this?

@IvanARashid and @siria98; we would have to play around with how to run these unit tests... I think currently use stuff from Zenodo in my volume testing already. So I think it is possible, as long as you run the get zenodo code before using it.

